PR #2, opened 2016-12-13
Enable Terraform remote state with Azure Blob storage

rtyler:
See [INFRA-996](https://issues.jenkins-ci.org/browse/INFRA-996)


olblak:
lgtm01

PR #5, opened 2016-12-15
K8s/01

olblak:
- Add a plan to build k8s
- Add a tf plan to build docker registry
- Add location variable

Feel free to comment 
 - location variable
 - docker registry name


plans/dockerregistry.tf
```diff
@@ -0,0 +1,119 @@
...
+    env = "${var.prefix}"
+  }
+}
+
+resource "azurerm_storage_account" "dockerregistry" {
+    name                = "${var.prefix}dockerregistry"
+    resource_group_name = "${azurerm_resource_group.dockerregistry.name}"
+    location            = "${var.location}"
+    account_type        = "Standard_GRS"
+    tags {
```
rtyler: I wonder if this is something that we can apply to all resources

plans/dockerregistry.tf
```diff
@@ -0,0 +1,119 @@
...
+  parameters = {
+	registryName = "${var.prefix}registry"
+ 	registryLocation = "${var.location}"
+	registryApiVersion = "2016-06-27-preview"
+	storageAccountName = "${ azurerm_storage_account.dockerregistry.name }"
+	storageAccountType = "${ azurerm_storage_account.dockerregistry.account_type }"
+	#adminUserEnabled = true	
+  }
+  deployment_mode = "Incremental"
+  template_body = <<DEPLOY
```
rtyler: I think to make our lives easier, all our ARM templates should be put into a separate directory, perhaps `./arm-templates` in the root directory?

plans/dockerregistry.tf
```diff
@@ -0,0 +1,119 @@
...
+    name                = "${var.prefix}dockerregistry"
+    resource_group_name = "${azurerm_resource_group.dockerregistry.name}"
+    location            = "${var.location}"
+    account_type        = "Standard_GRS"
+    tags {
+        env = "${var.prefix}"
+    }
+}
+
+resource "azurerm_template_deployment" "dockerregistry"{
```
rtyler: Per our in-person discussion, this should use the [depends_on](https://www.terraform.io/docs/configuration/resources.html#explicit-dependencies) parameter to explicitly depend on the resources defined above such that they are created before the template is submitted for deployment

plans/k8s.tf
```diff
@@ -0,0 +1,201 @@
...
+    tags {
+        env = "${var.prefix}"
+    }
+}
+
+resource "azurerm_template_deployment" "k8s"{
+  name  = "${var.prefix}k8s"
+  resource_group_name = "${ azurerm_resource_group.k8s.name }"
+  parameters = {
+        sshRSAPublicKey = "${file("~/.ssh/id_rsa.pub")}"
```
rtyler: This isn't going to work when run inside of ci.jenkins.io or trusted.ci :frowning:

plans/k8s.tf
```diff
@@ -0,0 +1,201 @@
...
+        dnsNamePrefix = "${var.prefix}"
+# Bug with integer
+#        agentCount = 1
+#        masterCount = 1
+        agentVMSize = "Standard_A2"
```
rtyler: I think this should be a variable like `k8_agent_size` so production environments can be much bigger VMs

plans/variables.tf
```diff
@@ -3,3 +3,4 @@ variable "client_id" {}
 variable "client_secret" {}
 variable "tenant_id" {}
 variable "prefix" {}
+variable "location" {}
```
rtyler: I have a similar commit which defines this, I think the variable should define a default, e.g.

```
variable "location" {
    type    = "string"
    default = "East US 2"
}
```

olblak: Fully agree, I was doing that initially

olblak: but we shouldn't use east us 2 as it doesn't support all azure services

rtyler:
Some of this work is going to depend on [INFRA-1002](https://issues.jenkins-ci.org/browse/INFRA-1002) being done, which is currently on my plate.

plans/k8s.tf
olblak: We should be able to define ssh_pubkey_path

plans/dockerregistry.tf
olblak: ack

plans/k8s.tf
olblak: -> https://azure.microsoft.com/en-us/regions/services/

rtyler:
@olblak I tested this out locally insofar that I was able to provision a full k8s cluster from the terraform in this pull request. I didn't have the time to do much else though, is there something more focused that you would like me to test?

olblak:
You may want to try the docker registry.

olblak:
Btw the check fail because I changed the default region to east us where azure have more services.
So before the check pass we must delete all old ressources from East US 2 (related to PR5)

plans/dockerregistry.tf
```diff
@@ -0,0 +1,33 @@
...
+  name  = "${var.prefix}dockerregistry"
+  resource_group_name = "${ azurerm_resource_group.dockerregistry.name }"
+  depends_on          = ["azurerm_resource_group.dockerregistry"]
+  parameters = {
+	registryName = "${var.prefix}registry"
+ 	registryLocation = "${var.dockerregistrylocation}"
+	registryApiVersion = "2016-06-27-preview"
+	storageAccountName = "${ azurerm_storage_account.dockerregistry.name }"
+	#adminUserEnabled = true
```
rtyler: Does this imply that the we will be unable to administer the configured registry?

plans/dockerregistry.tf
```diff
@@ -0,0 +1,33 @@
...
+  resource_group_name = "${ azurerm_resource_group.dockerregistry.name }"
+  depends_on          = ["azurerm_resource_group.dockerregistry"]
+  parameters = {
+	registryName = "${var.prefix}registry"
+ 	registryLocation = "${var.dockerregistrylocation}"
+	registryApiVersion = "2016-06-27-preview"
+	storageAccountName = "${ azurerm_storage_account.dockerregistry.name }"
+	#adminUserEnabled = true	
+  }
+  deployment_mode = "Incremental"
```
rtyler: Doesn't this mean we will not be idempotently creating registries?

plans/k8s.tf
```diff
@@ -0,0 +1,38 @@
...
+}
+
+resource "azurerm_template_deployment" "k8s"{
+  name  = "${var.prefix}k8s"
+  resource_group_name = "${ azurerm_resource_group.k8s.name }"
+  depends_on          = ["azurerm_resource_group.k8s"]
+  parameters = {
+        sshRSAPublicKey = "${var.ssh_pubkey}"
+        dnsNamePrefix = "${var.prefix}"
+# Bug with integer
```
rtyler: Have you filed this yet?

plans/variables.tf
```diff
@@ -3,13 +3,25 @@ variable "client_id" {}
...
 }
-
 # Port used for Puppet agents to connect to a Puppet master
 variable "puppet_master_port" {
     default = 8140
 }
+# Define kubernetes agent instance size
+variable "k8s_agent_size" {
+    type = "string"
+    default = "Standard_A2"
```
rtyler: This is too small IMHO for production usage, D2v2 would be suitable I think

plans/variables.tf
```diff
@@ -3,13 +3,25 @@ variable "client_id" {}
...
+}
+# Define default ssh public key used to provision new kubernetes agent
+variable "ssh_pubkey" {
+    type = "string"
+    default = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC4Bvn9TvicSk/fBKHqMqSyR95FbPV+Ie2TJS9grKO7UByHErsSM5Pz1i9DO4l0LDma8vIaj/zn1RIN38Q96xwWd7dfWalmvZm3/hX5z5UW/KiVfla/BmltFZFwCtl6XjbDEtnkLMp+pMDzpbA8tbLOI23SQAj0R5XlmKGV8NluKGhFCTLQYWNIjoCdpYITOP7ibLDd3C9TzL1pucgbBG8R+/c1ijDkRhPUNj4s4jjawQ4rnSAj5Hm7jHQ93Bg2TiYjp3DoEGBtNjQoVbF+0ZbGQOydHj0QMhsle83nEjfdCahg21BquOW7kHV0gxPQs4LlpISKGhZ96yFqIgStUu2Z example@sshpublickey"
```
rtyler: This relates to [INFRA-1002](https://issues.jenkins-ci.org/browse/INFRA-1002) yes? I believe it might make sense to add a common public/private key for testing into this repository  (e.g. `keys/dummy-k8s` and `keys/dummy-k8s.pub`) so that test infrastructure can be connected to by anybody working with this repository.

Then when INFRA-1002 is finished, Jenkins can override this variable in trusted.ci for non-ephemeral clusters.

olblak: Nope, in fact I put a random ssh key, because we can't create a container service without a ''validate' ssh key.
This value is only for testing purpose.

rtyler:
@olblak Please rebase this again and reopen (you can just hit the "Re-open pull request" button after pushing).

I found the `Jenkinsfile` issue and fixed it in master, until pull requests rebase they will waste lots of Azure time, see [JENKINS-40722](https://issues.jenkins-ci.org/browse/JENKINS-40722) for more.

PR #8, opened 2016-12-15
Avoid parallelism in the bootstrapping for remote state

rtyler:
In theory, @olblak believes, this should help us avoid scenarios where the
Azure storage account which Terraform is going to be configured to use does not
properly exist by the time remote state is pushed.

Also extending the wait time to give Azure an extra 30 seconds to register the
remote state storage account


olblak:
lgtm

PR #10, opened 2016-12-26
Add terraform plans to provision logs resources

olblak:
Those resources are use to store various logs on ${PREFIX}logs
Right now we only use it to store kubernetes logs.

This commit add followings resources creation
	* resource groupe ${PREFIX}logs
	* storage account ${PREFIX}logs
	* disk shared storage ${PREFIX}logs
	* a OMS workspace link to disk shared storage ${PREFIX}logs

### DIFF
--- a/plans/logs.tf
+++ b/plans/logs.tf
@@ -1,0 +1,42 @@
+resource "azurerm_resource_group" "logs" {
+    name     = "${var.prefix}logs"
+    location = "${var.location}"
+    tags {
+        env = "${var.prefix}"
+    }
+}
+
+resource "azurerm_storage_account" "logs" {
+    name                = "${var.prefix}logs"
+    resource_group_name = "${azurerm_resource_group.logs.name}"
+    location            = "${var.location}"
+    account_type        = "Standard_GRS"
+    depends_on          = ["azurerm_resource_group.logs"]
+    tags {
+        env = "${var.prefix}"
+    }
+}
+
+resource "azurerm_storage_share" "logs" {
+    name = "logs"
+    resource_group_name     = "${azurerm_resource_group.logs.name}"
+    storage_account_name    = "${azurerm_storage_account.logs.name}"
+    quota                   = 50
+    depends_on              = ["azurerm_resource_group.logs","azurerm_storage_account.logs"]
+}
+
+resource "azurerm_template_deployment" "logs"{
+    name  = "${var.prefix}logs"
+    resource_group_name = "${ azurerm_resource_group.logs.name }"
+    depends_on          = ["azurerm_resource_group.logs"]
+    parameters = {
+        omsWorkspaceName = "${var.prefix}logs"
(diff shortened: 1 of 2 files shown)


plans/logs.tf
```diff
@@ -0,0 +1,42 @@
...
+resource "azurerm_template_deployment" "logs"{
+    name  = "${var.prefix}logs"
+    resource_group_name = "${ azurerm_resource_group.logs.name }"
+    depends_on          = ["azurerm_resource_group.logs"]
+    parameters = {
+        omsWorkspaceName = "${var.prefix}logs"
+        omsRegion = "${var.location}"
+        existingStorageAccountName = "${azurerm_storage_account.logs.name}"
+        existingStorageAccountResourceGroupName = "${azurerm_resource_group.logs.name}"
+        #table = ""
```
rtyler: ?

plans/logs.tf
```diff
@@ -0,0 +1,42 @@
...
+    }
+    #deployment_mode = "Complete"
+    deployment_mode = "Incremental"
```
rtyler: Before this is merged, shouldn't "Incremental" be removed?

I'm assuming that with "Incremental" the ARM template is not idempotent?

plans/logs.tf
```diff
@@ -0,0 +1,42 @@
...
+    tags {
+        env = "${var.prefix}"
+    }
+}
+
+resource "azurerm_storage_account" "logs" {
+    name                = "${var.prefix}logs"
+    resource_group_name = "${azurerm_resource_group.logs.name}"
+    location            = "${var.location}"
+    account_type        = "Standard_GRS"
```
rtyler: Are there size concerns we should have here? Assuming we're dumping all logs into this storage account.

rtyler:
Please rebase this off master, it's missing some `Jenkinsfile` updates that are causing the Pipeline to fail in Jenkins

rtyler:
Re-open when it's rebased plz

plans/logs.tf
olblak: Indeed, I think that using 50GB is fine for now but we may want to increase the size

plans/logs.tf
olblak: If I understand correctly terraform documentation, we must use incremental
Otherwise we will destroy other resources. 

_deployment_mode - (Required) Specifies the mode that is used to deploy resources. This value could be either Incremental or Complete. Note that you will almost always want this to be set to Incremental otherwise the deployment will destroy all infrastructure not specified within the template, and Terraform will not be aware of this._

plans/logs.tf
rtyler: Can we upgrade the storage account size in place? That would be good to answer before this goes to production

plans/logs.tf
rtyler: Understood. Please ensure that this Terraform resource behaves in an idempotent manner nonetheless

PR #14, opened 2017-01-04
[INFRA-1015] Logs resources

olblak:
This PR add followings resources creation

- resource groupe ${PREFIX}logs
- storage account ${PREFIX}logs
- OMS workspace linked '${PREFIX}logs' storage account

Create Log analytics resources to display logs generated by kubernetes + fluentd.
Following resources were tested by this image -> https://github.com/olblak/fluentd-k8s-azure
On a kubernetes azure container service.

Log architecture work as follow:
|Kubernetes | -> |Fluentd| -> |Log Analytics|

N.B.: With this approach we don't use blob storage

### DIFF
--- a/arm_templates/logs.json
+++ b/arm_templates/logs.json
@@ -1,0 +1,109 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "omsWorkspaceName": {
+      "type": "string",
+      "defaultValue": "omsws01",
+      "metadata": {
+        "description": "Specify name of your OMS workspace"
+      }
+    },
+    "omsRegion": {
+      "type": "string",
+      "defaultValue": "West Europe",
+      "allowedValues": [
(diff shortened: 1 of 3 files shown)


rtyler:
> Log architecture work as follow:
> |Kubernetes | -> |Fluentd| -> |Log Analytics|
>
> N.B.: With this approach we don't use blob storage

@olblak Doesn't this mean that the logs will only be searchable/readable for a limited time? From my understanding of OMS it would only store the last 1-30 days of log information? For some types of logs, like access logs which can be used for deriving helpful metrics, this is not ideal :frowning_man: 

I don't think we discussed log retention policies much before, so I'm curious if this "direct-to-OMS" approach is going to cause issues with that.

olblak:
Indeed it's a question that we must thing about. 
Here is the pricing https://azure.microsoft.com/en-us/pricing/details/log-analytics/
It's free for 7days and if we want more, we must pay.

We can decide to ship logs in two places 
-> To log analytics (fluent plugin)
-> To a storage account.
      -> To a shared disk, configured at Kubernetes level (I

PR #15, opened 2017-01-04
K8s/01

olblak:
Add a plan to build azure container service (kubernetes)
Add a tf plan to build docker registry
Add location variable


olblak:
I can't find how to reopen a closed PR so I am not sure that closing PR each time is to best way to go

olblak:
Regarding two previous comments.

1. Instance size, I update it to d2_v2
2. Default ssh key, I prefer to define a random ssh key and override the value from puppet

rtyler:
> Default ssh key, I prefer to define a random ssh key and override the value from puppet

I think there may have been a misunderstanding in what I was asking for. I want a dummy SSH created and *committed* to this repository. Because I (or any other contributor) do not have access to the private key which corresponds to the public key in the Terraform code, I would need to change code in order to connect to the K8s cluster.

[This is effectively what I am suggesting](https://github.com/mitchellh/vagrant/tree/master/keys#insecure-keypair)

olblak:
It was already a fake key based on a real one.
I just changed it to a more obvious fake key.

rtyler:
@olblak I think you're missing my point. With this pull request, any contributor will have to update or override their variables in order to have access to their provisioned k8s cluster.

I was suggesting following the Vagrant approach of including insecure SSH keys, for testing only, in the repository for ease-of-use.

olblak:
@rtyler Indeed I was missing your point.
It's a good idea, I added a dummy private/public ssh key

olblak:
Yes, let's merge it

PR #16, opened 2017-01-07
Change our default ARM template format to YAML to allow for inline commenting

rtyler:
@olblak I wanted to try this out to see if maintaining YAML-based ARM templates would be any nicer than the awful JSON, and I have concluded that this makes life *much* easier as ARM template complexity grows.

This was suggested by a former colleague @jotto: https://twitter.com/jotto/status/817546924517654528

Basically with the yaml2json script and some GNU/Make shenanigans all the yaml
files in arm_templates/ will be converted to JSON appropriately such that they
can still be referenced within the azurerm_template_deployment terraform
resource. E.g.

    esource "azurerm_template_deployment" "some_docdb" {
        name                = "just-another-nosql"
        resource_group_name = "${azurerm_resource_group.documentdb-test.name}"
        parameters          = {
            databaseAccountName = "my-documentdb-name"
        }
        deployment_mode     = "Incremental"
        template_body       = "${file("./arm_templates/documentdb.json")}"
    }


rtyler:
The build is failing because there isn't a `yaml` module in the Python environment. I'll figure that out while @olblak reviews this :)

olblak:
@rtyler It's a good a idea and easy to rollback if we are not happy with. so let's try

PR #18, opened 2017-01-23
Logs resources

olblak:
Add azure resources for kubernetes logging workflow.

- azure resource group
- azure storage account
- azure shared disk 
- OMS

We can use this [Project](https://github.com/olblak/k8s) to test this PR.
Keep in mind that we will have to configure secrets properly

N.B.: 
Due to some issues in [fluent-plugin-azurestorage](https://github.com/htgc/fluent-plugin-azurestorage/issues), I choose to use an azure shared disk that I mount at kubernetes level.
We should switch to an azure blob storage if the plugin work as expected.

[This docker image](https://github.com/olblak/fluentd-k8s-azure) implement the logging workflow explain 
[here - IEP4](https://github.com/jenkins-infra/iep/pull/5)  (logtype: archive, logtype:stream,...)


olblak:
Btw if you want to see logs from log analytics, you need to enable 'custom logs' which is a preview feature that need to be enable manually through web interface

rtyler:
@olblak The "Custom Logs" feature, can you document where that is? Is that a subscription-wide feature? If so, I can enable it for the whole Jenkins subscription

rtyler:
Looks like the Pipeline run did succeed but failed to update commit status: `Could not update commit status. Message: GitHub API rate limit exceeded`

arm_templates/logs.json
```diff
@@ -0,0 +1,109 @@
+{
```
rtyler: This should be YAML now, in fact the `.gitignore` file is set up to ignore all JSON in t his directory.

Once converted to yaml you can add some comments like where it's from, etc :smile:

olblak:
I add two screenshots on this [repository](https://github.com/olblak/k8s/tree/master/doc)
Should I add them here? in IEP?

olblak:
I am closing this PR as commits' history got polluted.
So I prefer to create a cleaner branch which can be merge from [here](https://github.com/jenkins-infra/azure/pull/19)

PR #19, opened 2017-01-24
Add logs resources provisionning v2.0

olblak:
Add azure resources for kubernetes logging workflow.

	* Create logs resource group
	* Create logs storage account
	* Create logs shared disk storage
	* Create OMS workspace

	More informations can be found within this project
	-> [IEP-004](https://github.com/jenkins-infra/iep/pull/5)

We can use this [Project](https://github.com/olblak/k8s) to test this PR.
Keep in mind that we will have to configure secrets properly

N.B.: 
Due to some issues in [fluent-plugin-azurestorage](https://github.com/htgc/fluent-plugin-azurestorage/issues), I choose to use an azure shared disk that I mount at kubernetes level.
We should switch to an azure blob storage if the plugin work as expected.

[This docker image](https://github.com/olblak/fluentd-k8s-azure) implement the logging workflow explain 
[here - IEP4](https://github.com/jenkins-infra/iep/pull/5)  (logtype: archive, logtype:stream,...)

I add two screenshots on this [repository](https://github.com/olblak/k8s/tree/master/doc)
-> [enable_azure_custom_logs](https://github.com/olblak/k8s/blob/master/doc/enable_azure_custom_logs.png)
Should we add them here? in IEP?

### DIFF
--- a/arm_templates/logs.yaml
+++ b/arm_templates/logs.yaml
@@ -1,0 +1,81 @@
+$schema: https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json
+contentVersion: 1.0.0.0
+parameters:
+  omsWorkspaceName:
+    type: string
+    defaultValue: omsws01
+    metadata:
+      description: Specify name of your OMS workspace
+  omsRegion:
+    type: string
+    defaultValue: West Europe
+    allowedValues:
+    - West Europe
+    - East US
+    - Southeast Asia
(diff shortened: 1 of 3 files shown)


rtyler:
@olblak FYI, this  failed to deploy in production because the storage account "prodlogs" was already taken. I didn't pick up on this during review but we need to always have unique strings in our storage account names (probably containing "jenkins")

PR #22, opened 2017-02-08
Scripts updates

rtyler:


scripts/kubectl
```diff
@@ -20,11 +20,11 @@ fi;
 # Passing LOGNAME and USER because the acs component in the azure-cli does some
 # user lookup
 exec docker run --rm  -u $(id -u):$(id -u) \
-    -ti \
     -v ${PWD}/.kube:/.kube \
     -v ${REPO_DIR}/ssh_key:/.ssh \
     -e LOGNAME=$LOGNAME \
     -e USER=$USER \
     -v $PWD:/data \
+    -p 8001:8001 \
```
olblak: Why do we need to open port 8001?

rtyler: For the proxy subcommand

rtyler:
> wouldn't be better to use ${HOME}/.kube instead of ${PWD}/.kube ?

@olblak the point of using `$PWD` is that the project directory will contain the `.kube/config` so there is _not_ a global configuration and instead things live in the project directory.

scripts/kubectl
olblak: make sense

olblak:
Ok I am fine with that approach if we agree that we only share "$PWD/jenkins-infra/azure/" project directory within docker image.
That means if a docker container need to create files, they will only be in "$PWD/.."

PR #23, opened 2017-04-26
Pin our terraform container to 0.8.8

rtyler:
The latest 0.9.x doesn't support `terraform remote`


rtyler:
The PR is failing because it must first checkout master (which is broken) in order to provision the master branch of terraform plans.

I'm going to merge this to fix master regardless

PR #25, opened 2017-05-10
Add shared storage 'accountapp' needed to store election logs

olblak:
Add shared storage 'accountapp'
This shared storage is needed by the accountapp to store election logs.

Linked to following PR:
[Jenkins Infra - #739](https://github.com/jenkins-infra/jenkins-infra/pull/739)
[Accountapp - #104](https://github.com/jenkins-infra/account-app/pull/104)

### DIFF
--- a/plans/accountapp.tf
+++ b/plans/accountapp.tf
@@ -1,0 +1,29 @@
+#
+# This terraform plan defines the resources necessary to host accounts.jenkins.io files
+#
+resource "azurerm_resource_group" "accountapp" {
+    name     = "${var.prefix}accountapp"
+    location = "${var.location}"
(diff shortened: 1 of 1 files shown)


plans/logs.tf
```diff
@@ -26,6 +26,14 @@ resource "azurerm_storage_share" "logs" {
     depends_on              = ["azurerm_resource_group.logs","azurerm_storage_account.logs"]
 }
 
+resource "azurerm_storage_share" "accountapp" {
```
rtyler: Why is this in the logs plan? I think this should be in a plan all to itself for the account app in case we need to provision new resources later.

PR #27, opened 2017-06-09
Add some Azure resources for Azure-hosted backups

rtyler:
See INFRA-1148

See also jenkins-infra/jenkins-infra#764

### DIFF
--- a/plans/backups.tf
+++ b/plans/backups.tf
@@ -1,0 +1,35 @@
+#
+# This terraform plan defines the resources necessary to store varied backups
+# for Jenkins hosts to publish to
+#
+# See: https://issues.jenkins-ci.org/browse/INFRA-1148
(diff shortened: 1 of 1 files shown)


rtyler:
@olblak Right, I was mis-remembering which level in Azure the access keys applied to. I'll rework this into two storage accounts.

olblak:
Looks like that terraform is not happy

* azurerm_storage_container.private_backups: only lowercase alphanumeric characters and hyphens allowed in "name": "private_backups"

* azurerm_storage_container.public_backups: only lowercase alphanumeric characters and hyphens allowed in "name": "public_backups"

rtyler:
Huh, I don't remember that being a problem before. Perhaps there's a validation issue happening now.

PR #28, opened 2017-06-16
Move the logs storage account into the same region as Kubernetes

rtyler:
After some triaging by @olblak, it appears that file volumes cannot be mounted
in k8s unless the storage account is in the same region as the k8s cluster.

This commit forces the creation of a new resource, which as @olblak and I
discussed in the ticket, is acceptable in this case since nothing is using this
storage account yet

Fixes INFRA-1186

### DIFF
--- a/plans/logs.tf
+++ b/plans/logs.tf
@@ -1,6 +1,6 @@
 resource "azurerm_resource_group" "logs" {
     name     = "${var.prefix}logs"
-    location = "${var.logslocation}"
+    location = "${var.location}"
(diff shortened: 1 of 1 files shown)


rtyler:
`Apply complete! Resources: 2 added, 5 changed, 2 destroyed.`

rtyler:
It looks like there are some weirdnesses with Azure and Terraform right now. I'm purging all the left-over pull request related resource groups in our Azure subscription and then will retry.

PR #31, opened 2017-07-07
Add a new ACS cluster

olblak:
Add new ACS cluster.
I explicitly define a resource name different from k8s to avoid confusion and it seems that someone prefer  vegetable over cakes. 

In order to avoid to pay for unused resources, this PR should be merge when we decide to migrate the Kubernetes cluster

### DIFF
--- a/plans/pea.tf
+++ b/plans/pea.tf
@@ -1,0 +1,59 @@
+resource "azurerm_resource_group" "pea" {
+  name     = "${var.prefix}pea"
+  location = "${var.location}"
+  tags {
+    env = "${var.prefix}"
+  }
+}
+
+resource "azurerm_storage_account" "pea" {
+    name                = "${azurerm_resource_group.pea.name}"
+    resource_group_name = "${azurerm_resource_group.pea.name}"
+    location            = "${var.location}"
+    account_type        = "Standard_GRS"
+    depends_on          = ["azurerm_resource_group.pea"]
+    tags {
+        env = "${var.prefix}"
+    }
+}
+
+resource "azurerm_container_service" "pea" {
+  depends_on             = ["azurerm_resource_group.pea"]
+  name                   = "${azurerm_resource_group.pea.name}"
+  location               = "${azurerm_resource_group.pea.location}"
+  resource_group_name    = "${azurerm_resource_group.pea.name}"
+  orchestration_platform = "Kubernetes"
+
+  master_profile {
+    count      = 3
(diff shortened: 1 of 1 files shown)


plans/pea.tf
```diff
@@ -0,0 +1,60 @@
...
+resource "azurerm_container_service" "pea" {
+  depends_on             = ["azurerm_resource_group.pea"]
+  name                   = "${var.prefix}pea"
+  location               = "${azurerm_resource_group.pea.location}"
+  resource_group_name    = "${azurerm_resource_group.pea.name}"
+  orchestration_platform = "Kubernetes"
+
+  master_profile {
+    count      = 1
+    dns_prefix = "${var.prefix}pea"
```
rtyler: I don't see anything here about the Master size? We need a master with a bigger disk to avoid some of the issues we have had thus far

plans/pea.tf
```diff
@@ -0,0 +1,60 @@
...
+    admin_username = "azureadmin"
+
+    ssh_key {
+      key_data = "${file("${var.ssh_pubkey_path}")}"
+    }
+  }
+
+  agent_pool_profile {
+    name       = "default"
+    count      = 1
```
rtyler: Only one agent in the agent pool seems very low to me, at least 3 seems right.

plans/pea.tf
olblak: We can't customize the size of the disk.
It's always 30GB for the root partition
But the bug we had with the current cluster is fixed in version > 1.5

plans/pea.tf
olblak: We can have 1 - 3 - 5 master and 1 to 100 agent.
I suggest to configure 3 master and 2 agent

olblak:
It seems that this PR is affected by the same bug than #27. 
Terraform timeout when it tries to provision the k8s cluster.
A solution would be to only delete azure resource if the build succeed.

plans/pea.tf
rtyler: Why more masters than agents? Also, can we configure the agents to be elastic so the number continues to grow?

olblak: Agent autoscaling is not yet supported but we can easily add/remove agent up to 100 agent
Number of Master can only be 1, 3, 5 and can't be modified after cluster creation (I need to validate this).

rtyler: Let's do 3 and 3 then :smile:

olblak:
I clean up this branch so it only add the new kubernetes cluster (pea)

PR #32, opened 2017-07-25
[WIP] [INFRA-1288] Upgrade terraform to 0.9.9

olblak:
! This PR upgrade terraform to version 0.9 and require manual steps to migrate remote state to backend as explain [here](https://www.terraform.io/docs/backends/legacy-0-8.html)

Remark: Manual step is done in stage('Pull remote') which should be remove once the backend is upgraded

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -51,7 +51,24 @@ try {
             node('docker') {
                 deleteDir()
                 git 'https://github.com/jenkins-infra/azure.git'
+                /* Create an empty terraform variables file so that everything can
+                 * be overridden in the environment
+                 */
                 sh "echo '{\"prefix\":\"${tfPrefix}\"}' > ${tfVarFile}"
+
+                /*
+                 * Three following actions must be remove once terraform is greater than 0.8
+                 * Cfr INFRA-1288
+                 */
+
+                sh 'rm plans/k8s.tf'
+                sh 'rm plans/logs.tf'
+                sh 'rm plans/dockerregistry.tf'
+
+                /*
+                 *INFRA-1288 end
+                 */
+
                 tfsh {
                     sh 'make deploy'
                 }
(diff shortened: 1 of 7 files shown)


Jenkinsfile
```diff
@@ -51,27 +51,57 @@ try {
...
                 deleteDir()
                 git 'https://github.com/jenkins-infra/azure.git'
+                /* Create an empty terraform variables file so that everything can
+                 * be overridden in the environment
+                 */
                 sh "echo '{\"prefix\":\"${tfPrefix}\"}' > ${tfVarFile}"
+
+                /*
+                 * Three following actions must be remove once terraform is greater than 0.8
+                 * Cfr INFRA-1288
```
rtyler: wtf does cfr mean :smiling_imp: 

But seriously, I don't understand why this is here, since this PR ugprades to a Terraform greater than 0.8

backend.tf
```diff
@@ -0,0 +1,3 @@
+terraform {
```
rtyler: This file shouldn't be here

Jenkinsfile
olblak: cfr is equivalent to cf (https://en.wikipedia.org/wiki/Cf.)
And seriously, in terraform 0.9, `terraform init` expect to be run at the root of the project where everything is located (plans,variables,...), and it also need a backend configuration file if we use another backend than local.
But for an unkowned reason, running Terraform 'operations' on subdirectories (like terraform apply plans) also complains for this missing backend configuration file.
This is why I define it in two places (root && plans), another solution would have been to reorganise the project structure (move all plans at the root) but in terraform 0.10 it should be possible to init subdirectories.

backend.tf
olblak: In terraform 0.9, `terraform init` expect to be run at the root of the project where everything is located (plans,variables,...), and it also need a backend configuration file if we use another backend than local.
But for an unkowned reason, running Terraform 'operations' on subdirectories (like terraform apply plans) also complains for this missing backend configuration file.
This is why I define it in two places (root && plans), another solution would have been to reorganise the project structure (move all plans at the root) but in terraform 0.10 it should be possible to init subdirectories.

Jenkinsfile
olblak: Terraform 0.9 introduce a new backend mechanism and the remote state must be pull with therraform client 0.8 then reinitialize with terraform client version 0.9.
Once jenkins-infra/azure:master is updated to 0.9 and deployed correctly, we can remove those lines

PR #35, opened 2017-09-02
Delete old k8s resource

olblak:


plans/pea.tf
```diff
@@ -0,0 +1,59 @@
+resource "azurerm_resource_group" "pea" {
```
rtyler: I'm not clear why this is showing up as added. This plan already exists o_O

Perhaps rebase off the master branch again?

olblak: Indeed I did a bad rebase but it's fixed now

PR #36, opened 2017-10-20
[INFRA-1352] File storage resources for trustedci on azure

olblak:


olblak:
Yep it will be used for the jenkins home and shared between jenkins container.
Do you have any objections?

rtyler:
@olblak no, but that should be commented in the file

PR #45, opened 2017-12-12
[INFRA-1131] Add azure shared storage for ldap

olblak:
After reevaluation, this storage will only be used to store ldap backups

### DIFF
--- a/plans/ldap.tf
+++ b/plans/ldap.tf
@@ -25,6 +25,6 @@ resource "azurerm_storage_share" "ldap" {
     name = "ldap"
     resource_group_name     = "${azurerm_resource_group.ldap.name}"
     storage_account_name    = "${azurerm_storage_account.ldap.name}"
-    quota                   = 10
+    quota                   = 100
     depends_on              = ["azurerm_resource_group.ldap","azurerm_storage_account.ldap"]
 }
--- a/plans/ldap.tf
+++ b/plans/ldap.tf
@@ -1,0 +1,30 @@
+#
+# This terraform plan defines the resources necessary to host accounts.jenkins.io files
+#
+resource "azurerm_resource_group" "ldap" {
+    name     = "${var.prefix}ldap"
+    location = "${var.location}"
+    tags {
+        env = "${var.prefix}"
+    }
+}
+
+resource "azurerm_storage_account" "ldap" {
+    name                      = "${var.prefix}ldap"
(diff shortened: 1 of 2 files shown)


plans/ldap.tf
```diff
@@ -0,0 +1,30 @@
...
+        env = "${var.prefix}"
+    }
+}
+
+resource "azurerm_storage_share" "ldap" {
+    name = "ldap"
+    resource_group_name     = "${azurerm_resource_group.ldap.name}"
+    storage_account_name    = "${azurerm_storage_account.ldap.name}"
+    quota                   = 10
+    depends_on              = ["azurerm_resource_group.ldap","azurerm_storage_account.ldap"]
```
rtyler: This is also not necessary to be explicit, fwiw

plans/ldap.tf
```diff
@@ -0,0 +1,30 @@
...
+    tags {
+        env = "${var.prefix}"
+    }
+}
+
+resource "azurerm_storage_share" "ldap" {
+    name = "ldap"
+    resource_group_name     = "${azurerm_resource_group.ldap.name}"
+    storage_account_name    = "${azurerm_storage_account.ldap.name}"
+    quota                   = 10
```
rtyler: Right now we're not aggressively rotating the LDAP transaction logs which can be very very large. I'm assuming that this is coming over, so this `quota` might be too low.

plans/ldap.tf
```diff
@@ -0,0 +1,30 @@
...
+    }
+}
+
+resource "azurerm_storage_account" "ldap" {
+    name                      = "${var.prefix}ldap"
+    resource_group_name       = "${azurerm_resource_group.ldap.name}"
+    location                  = "${var.location}"
+    account_tier              = "Standard"
+    account_replication_type  = "GRS"
+    depends_on                = ["azurerm_resource_group.ldap"]
```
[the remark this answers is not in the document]
olblak: Indeed but I usually prefer to be more explicit as implicit rules are sometimes changed from a version to an other one :)

plans/ldap.tf
olblak: Currently the database size is 10Mb so I though that I was already large enough but I can increase this value, more space

olblak:
@rtyler Did you have time to (re)-review this?

rtyler:
I had not @olblak, I figured I had time since you were on vacation :smiling_imp:

PR #46, opened 2018-01-03
Avoid trying to use a custom domain outside of production

rtyler:


rtyler:
Gah, I'm pretty sure this is just due to credentials expiring in AAD

rtyler:
The way the PR Pipeline works, this is going to fail since it first ries to deploy `master`. I'm going to merge this.

PR #47, opened 2018-01-06
Rename trusted-agent-1 to -2 to force terraform to fully delete resources

rtyler:
Unfortunately the subnet resources cannot be swapped out otherwise. Easier to
just nuke it all from orbit


olblak:
Not needed anymore

PR #52, opened 2018-01-26
Propose small spelling correction to script string

jeis2497052:
should the string read "Proceed" ?
Hope this helps?


olblak:
Thanks for this

PR #54, opened 2018-02-25
Properly enable storage encryption for tfstate

rtyler:
This is enabled by default for new storage accounts, which led to failures when
running terraform locally.

Splitting separately from my other work.


rtyler:
@olblak if you have a chance, try this out before we merge it. The master branch fails because it creates a new tfstate storage account with encryption enabled, and then our Terraform tries to set it to unencrypted causing the error shown in Jenkins

olblak:
Lgtm

PR #62, opened 2018-09-15
Add uplink Database

olblak:
Deploy a postgresql database asked by @daniel-beck


olblak:
Then the name 'uplink' doesn't really represent the usage and 'utility' would be more appropriated, in fact I am a bit afraid to use the same database for different services as we may loose visibility on what's the real usage.

rtyler:
@olblak we don't have to use the same _database_, sorry, I misspoke.

We should use a "utility" database server, and that should have a uplink database within it, does that make sense. I just don't want to overprovision a database server for this little DB

PR #66, opened 2018-09-21
Adds a dynamic lock to prevent multiple plans/applies at the same time

mpailloncy:
To prevent multiple infrastructure changes at the same time on the
same environment, this change adds a lock on Jenkins' side.
All builds targeting the same environment/resource type will be applied
in sequential order.

NOTE: This change *requires* Lockable Resources Plugin to be installed.

cc @rtyler @olblak

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -39,96 +39,99 @@ else {
     tfPrefix = 'infraci'
 }
 
-try {
-    stage('Prepare') {
-        /* When planning and applying changes for a pull request, the Pipeline
-         * should first use the master branch which will create a remote state
-         * that can be continued from later, more accurately simulating an
-         * execution of terraform plans on an existing production
-         * infrastructure
-         */
-        if (env.CHANGE_ID) {
-            node('docker') {
-                deleteDir()
-                git 'https://github.com/jenkins-infra/azure.git'
-                /* Create an empty terraform variables file so that everything can
-                 * be overridden in the environment
-                 */
-                sh "echo '{\"prefix\":\"${tfPrefix}\"}' > ${tfVarFile}"
+// dynamic lock to prevent multiple PR/branches trying to plan/apply things at the same time to the same environment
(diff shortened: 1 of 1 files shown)


rtyler:
Based on my understanding of [how this is supposed to work](https://jenkins.io/blog/2016/10/16/stage-lock-milestone/) this seems reasonable

mpailloncy:
One inconvenient with this plugin is that it creates one "lock" for each branch/PR (because locks name are based on ${tfPrefix}".  So that'll lead to a bunch of locks displayed in Jenkins global configuration page. 
You'll probably need to do some cleanup regularly.

olblak:
@mpailloncy, tell me if I am wrong but this doesn't prevent PR1 and PR2 to be run at the same time, it only lock job for a specific PR.
and if it also means that we'll have to do some clean up manually "regularly", I am not convinced that the benefits counter balances the additional work

mpailloncy:
@olblak indeed, by using ${tfPrefix} to name the lock, that doesn't prevent PR1 and PR2 to be built at the same time. But IIUC, we don't really care as all resources are created using that ${tfPrefix}, so there is no chance that conflict happen in that case. Am I right?
Also, it not only locks builds for a specific PR, but also for all branches (including master branch which targets production environment). So that prevents accidental concurrent changes if multiple merge are performed on master at the same time.

Anyways, it depends of what we want. Instead of using ${tfPrefix}, we could just create only one lock named "PR" for all Pull-Requests and another for "prod" if we're on master (or based on current branch name).

BTW, is the "Disable concurrent build" property ticked on trusted.ci.jenkins.io for this job? I don't see it on Jenkinsfile, but depending on how the job is set up, it could be ticked manually.

IMO the cleanup thing is not a big deal, really. That's not a mandatory action. I just mentioned it because I know that it could be confusing to see a lot of old locks listed in Jenkins global configuration. 
But since I'll not be the one who will need to do the cleanup, my opinion is probably not helpful here ;)

PR #67, opened 2018-09-26
Replace dns record ending by . with the zone name

olblak:
It seems that azure dns doesn't follow the same rules than bind


plans/dns_jenkinsci.tf
```diff
@@ -138,7 +138,7 @@ resource "azurerm_dns_mx_record" "jenkinsci_root_mx_entries" {
 
   record {
     preference = 0
```
rtyler: This host simply doesn't exist anymore, not sure why we're bothering with it

PR #86, opened 2019-02-12
Azure DNS ended up fuxoring up the CNAMEs for DCV

rtyler:
The records as they were deployed ended up being
`_9ede1b274d351e6d534fdc94833df749.jenkins-ci.org.jenkins-ci.org` which is
definitely wrong =_=

See INFRA-2007


rtyler:
Hopping over CI, yolo :man_dancing:

PR #88, opened 2019-03-16
Add a ci.jenkins.io master VM

rtyler:
This is basically just copying the great work that @olblak already did for the
cert.ci instance.


rtyler:
I'm going to try to bring this into production. The issues in CI are completely unrelated and I'd like to give this a try in production :smile:

PR #89, opened 2019-03-18
Update ci.jenkins.io endpoint

olblak:
I think it safe to switch traffic on azure (once another rsync is done)


olblak:
I am going to copy ci.jenkins.io letsencrypt certificate to azure.ci.jenkins.io so https keep working. 
I have until 22 of May before that cert expired :)

olblak:
done

PR #90, opened 2019-03-25
[INFRA-1352] Provision a new azure vm for trusted.ci

olblak:


olblak:
I currently don't have the time to work on this so I close it for now

PR #125, opened 2019-12-14
Disable hns

timja:
build is failing due to trying to update a property that can't be updated

[2019-12-11T16:04:30.828Z] * azurerm_storage_account.tfstate_datadog: Error creating Azure Storage Account "infracitfstatedatadog": storage.AccountsClient#Create: Failure sending request: StatusCode=400 -- Original Error: Code="AccountPropertyCannotBeUpdated" Message="The property 'isHnsEnabled' was specified in the input, but it cannot be updated."


timja:
Still failed, not really sure how to fix it,
We had a similar issue at work, I think we ended up recreating the storage account.

PR #132, opened 2020-01-09
Add azure storage blob for plugins.jenkins.io INFRA-2414

halkeye:


plans/pluginshtml.tf
```diff
@@ -0,0 +1,32 @@
+#
+# This terraform plan defines the resources necessary to host plugins.jenkins.io
+#
+
+resource "azurerm_resource_group" "pluginshtml" {
+  name     = "${var.prefix}pluginshtml"
```
timja: Why not call it plugin site? Rather than html I checked the other resources and they all use their site name

olblak: Definitely , while pluginsite is not the best name, at least it's consistent

PR #138, opened 2020-02-13
Increase max number of linux node for publick8s

olblak:


olblak:
This PR align with a change applied manually

PR #144, opened 2020-03-05
Add two share storages: mirrors & archives 

olblak:
This pull request defined two DNS records and share storages needed by services running on Kubernetes

* Dns record `get.jenkins.io`  and azure file storage needed for [PR#122](https://github.com/jenkins-infra/charts/pull/122) 

* Dns record `archives.azure.jenkins.io` and azure file storage needed for [PR#130](https://github.com/jenkins-infra/charts/pull/130)

In both cases the data need to be pushed from [here](https://github.com/jenkins-infra/jenkins-infra/blob/8ecc23a590eacddb50e784aeb503f7b9af63d866/dist/profile/files/mirrorbrain/sync.sh#L49)
using **blobxfer**

### DIFF
--- a/plans/dns_jenkinsio.tf
+++ b/plans/dns_jenkinsio.tf
@@ -44,6 +44,7 @@ locals {
   jenkinsio_cname_records = {
     # Azure
     accounts = "public.aks.jenkins.io"
+    get      = "public.aks.jenkins.io"
     javadoc  = "public.aks.jenkins.io"
     plugins  = "public.aks.jenkins.io"
     reports  = "public.aks.jenkins.io"
--- a/plans/dns_jenkinsio.tf
+++ b/plans/dns_jenkinsio.tf
@@ -43,13 +43,14 @@ locals {
 
   jenkinsio_cname_records = {
     # Azure
-    accounts = "public.aks.jenkins.io"
-    get      = "public.aks.jenkins.io"
-    javadoc  = "public.aks.jenkins.io"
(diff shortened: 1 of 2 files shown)


plans/dns_jenkinsio.tf
```diff
@@ -43,12 +43,14 @@ locals {
...
   jenkinsio_cname_records = {
     # Azure
-    accounts = "public.aks.jenkins.io"
-    javadoc  = "public.aks.jenkins.io"
-    plugins  = "public.aks.jenkins.io"
-    reports  = "public.aks.jenkins.io"
-    www      = "jenkins.io"
-    uplink   = "public.aks.jenkins.io"
+    accounts       = "public.aks.jenkins.io"
+    archives.azure = "public.aks.jenkins.io"
```
timja: does this need .azure in it?

there's a dns record for archives.jenkins.io but it doesn't go anywhere just a blank apache page

olblak: Right now there is only [http://archives.jenkins-ci.org](http://archives.jenkins-ci.org/ )listening on http only

That machine is running on rackspace and cost us around 500$/month, I would like to run archives.azure.jenkins.io during a testing period before switching officially to the k8s version 

Archives.jenkins-ci.org is managed by puppet with this [role](https://github.com/jenkins-infra/jenkins-infra/blob/staging/dist/profile/manifests/archives.pp)

PR #149, opened 2020-04-16
Premium storage for releases

timja:


timja:
Bit late to change this now

PR #162, opened 2020-10-07
chore: update docs to retrieve the subscription id

garethjevans:


HACKING.adoc
```diff
@@ -58,15 +58,18 @@ Service Principle associated with the application and grant permissions to that.
 az ad sp create --id $(az ad app list -o tsv --query "[?displayName=='jenkins-terraform'].appId"`)
 ----
 
+Retrieve the subscriptionId
+[source]
+----
+SUBSCRIPTION_ID=$(az account list -o tsv --query "[?name=='Jenkins'].id")
```
timja, line 64: could also do:
```suggestion
SUBSCRIPTION_ID=$(az account show --query id -o tsv)
```

but the above might be safer

timja (approved):

markyjackson-taulia (approved):

HACKING.adoc
olblak: In fact, I like both :p, one when we work with Jenkins Azure account and the other one when work on a staging/dev account

olblak:
```suggestion
SUBSCRIPTION_ID=$(az account list -o tsv --query "[?name=='Jenkins'].id")
or 
SUBSCRIPTION_ID=$(az account show --query id -o tsv)
```

markyjackson-taulia (approved):

PR #168, opened 2021-01-11
Open rsyncd port on k8s cluster

olblak:
Signed-off-by: Olivier Vernin <olivier@vernin.me>

### DIFF
--- a/plans/vnets-nsg.tf
+++ b/plans/vnets-nsg.tf
@@ -86,6 +86,20 @@ resource "azurerm_network_security_rule" "public-app-tier-allow-ldaps-inbound" {
   network_security_group_name = azurerm_network_security_group.public_app_tier.name
 }
 
+resource "azurerm_network_security_rule" "public-app-tier-allow-rsync-inbound" {
+  name                        = "allow-ldaps-inbound"
+  priority                    = 102
+  direction                   = "inbound"
+  access                      = "allow"
+  protocol                    = "tcp"
+  source_port_range           = "*"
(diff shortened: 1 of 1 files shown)


plans/vnets-nsg.tf
```diff
@@ -86,6 +86,20 @@ resource "azurerm_network_security_rule" "public-app-tier-allow-ldaps-inbound" {
...
+resource "azurerm_network_security_rule" "public-app-tier-allow-rsync-inbound" {
+  name                        = "allow-rsyncd-inbound"
+  priority                    = 103
+  direction                   = "inbound"
+  access                      = "allow"
+  protocol                    = "tcp"
+  source_port_range           = "*"
+  destination_port_range      = "873"
+  source_address_prefix       = "*"
+  destination_address_prefix  = "*"
```
dduportal, line 98: Is there a possibility to restrict the destination (e.g. either the loadbalancer IP used by the rsyncd service, or the ingress' IP) ? This would avoid unwanted rsync traffic going to the networ'ks host right?

olblak: the destination is going to be annoying because we use ephemeral private IP network, but we should filter based on source IP address like we are doing at the loadbalancer level and image level, [here](https://github.com/jenkins-infra/charts/blob/master/charts/mirror/values.yaml#L42L43) and [there](https://github.com/jenkins-infra/docker-rsyncd/blob/2123bce4d9443ad6becc4b23a0a930bf6af57c97/config/rsyncd.conf#L30)

PR #170, opened 2021-04-19
Failover get.jenkins.io to pkg.origin.jenkins.io

timja:


dduportal:
@olblak @timja what is the status: should we close or revert this?

olblak:
I would like to do some tests on get.jenkins.io before reverting

PR #171, opened 2021-05-22
Add validation TXT entry for libera.chat

halkeye:
They emailed me data to get entered into the domain to allow claiming of ownership


timja (approved):
applied manually, not sure the state of this pipeline I'll leave it for @olblak / @dduportal

olblak (approved):
At the moment we still have to manually apply it

PR #178, opened 2022-03-24
feat(public-pgsql) add a postgresql flexible server instance for the public applications

dduportal:
Part of https://github.com/jenkins-infra/helpdesk/issues/1627 (migrating ratings.jenkins.io from AWS VM to Azure AKS).

This PR adds a postgresql flexible server for the public applications:

- with its own subnet since the azure documentation explains that it must be reserved ("delegated" subnet)
  - That subnet should not be reachable from the outside, and would only be allowed incoming connections from AKS subnet.
- Admin and password are generated by terraform (and stored in the encrypted backend state). Please note that the resources of kind `random_password` are considered as sensitive as per [its documentation](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password) and their output hidden from the console log.
- For sizing, we took the default recommended burstable instance size with 1 vCPU and 2 Gb, with a data storage of 20 Gb
    - Current rating database in AWS is an RDS on a t2.micro (1vCPU/1Gb) with 5Gb of data, which is ~55% full
    - This postgres won't be only used for rating in the future, hence the instance size with more RAM and more storage.
    - Burstable serie to ensure that we can outperform the CPU when there is a peak without paying too much. That might need to grow based on metric on the future.
    - Please note that the most recent LTS version of postgres is used: v13.x
    - Choice of flexible (instead of single) to ensure we are using latest azure recommendations. HA is not enabled for now but it can be at any moment.

Please note the followings:
- This is a first iteration that we already tested successfully with @smerle33 and we destroyed it to ensure that it is created and maintained by the full infra-as-code process
- No database created yet for rating.jenkins.io (2nd iteration)
- The connectivity from AKS will be tested and eventually updated after the initial creation
- For now, the public network is only referenced as a "data". It will have to be imported and managed soon.

### DIFF
--- a/vnets.tf
+++ b/vnets.tf
@@ -1,0 +1,55 @@
+#
+# This terraform plan defines the resources necessary to provision the Virtual
+# Networks in Azure according to IEP-002:
+#   <https://github.com/jenkins-infra/iep/tree/master/iep-002>
+#
+#                        +---------------------+
+#                        |                     |
+#      +---------------> |  Public Production  <-------+
+#      |                 |                     |       |
+#      |                 +---------------------+     VNet Peering
+#      |                                               |
+#      |                                 +-------------v--------+
+#                        +-------------+ |                      |
+# The Internet --------> + VPN Gateway |-|  Private Production  |
+#                        +-------------+ |                      |
+#      |                                 +----------------------+
+#      |
+#      |                 +----------------+
+#      |                 |                |
+#      +---------------> |   Development  |
+#                        |                |
+#                        +----------------+
+#
+## RESOURCE GROUPS
+################################################################################
+data "azurerm_resource_group" "public_prod" {
+  name = "prod-jenkins-public-prod"
+}
+
+################################################################################
+## VIRTUAL NETWORKS
+################################################################################
+data "azurerm_virtual_network" "public_prod" {
+  name                = "prod-jenkins-public-prod"
+  resource_group_name = data.azurerm_resource_group.public_prod.name
+  # address_space       = ["10.0.0.0/16"]
+}
+
+################################################################################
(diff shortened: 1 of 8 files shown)


vnets.tf
```diff
@@ -0,0 +1,55 @@
...
+#      |                 +----------------+
+#      |                 |                |
+#      +---------------> |   Development  |
+#                        |                |
+#                        +----------------+
+#
+## RESOURCE GROUPS
+################################################################################
+data "azurerm_resource_group" "public_prod" {
+  name = "prod-jenkins-public-prod"
```
lemeurherve, line 27: What's the purpose of the repetition of "prod"? How about "jenkins-public-prod"?

dduportal: It's a `data` in term of terraform, meaning that it is a reference to an existing resource. So we do not have the choice except using the actual name of the existing resource.

@smerle33 and I wondered the same and found the following explanations for this naming though:
- https://github.com/jenkins-infra/azure/blob/legacy-tf/plans/vnets.tf#L28 => there is a "prefix" variable,
- which has no default value: https://github.com/jenkins-infra/azure/blob/legacy-tf/plans/variables.tf#L13-L14.
- The value `prod` for this variable was defined by the pipeline (https://github.com/jenkins-infra/azure/blob/legacy-tf/Jenkinsfile#L26-L40) and passed to terraform through a variables file (https://github.com/jenkins-infra/azure/blob/legacy-tf/Jenkinsfile#L73)

=> We understood it was a way to separate prod / staging (named `infraci`) way before the terraform workdir. Difference with our current setup is that we define purely separated environments for prod and staging (better safe than sorry :D )

timja (approved):

PR #179, opened 2022-03-25
feat(rating): add the postgres database + users

smerle33:
following #178 

Part of jenkins-infra/helpdesk#1627 (migrating [ratings.jenkins.io](http://ratings.jenkins.io/) from AWS VM to Azure AKS).

using the resource provider cyrilgdn : https://registry.terraform.io/providers/cyrilgdn/postgresql/latest
cyrilgdn is a third party provider but is widely used, it's quite up to date https://github.com/cyrilgdn/terraform-provider-postgresql/blob/ec7110229606a3c2b175714ea94f780728973f4d/go.mod#L3 and https://github.com/cyrilgdn/terraform-provider-postgresql/releases

coauthored with @dduportal


locals.tf
```diff
@@ -0,0 +1,3 @@
+locals {
+  public_pgsql_admin_login = "psqladmin${random_password.pgsql_admin_login.result}"
+}
```
lemeurherve, line 3:
```suggestion
}

```

lemeurherve (approved):

lemeurherve (approved):

providers.tf
```diff
@@ -2,3 +2,11 @@
 provider "azurerm" {
   features {}
 }
+
+provider "postgresql" {
+  # Configuration options
+  host     = azurerm_postgresql_flexible_server.public.fqdn
+  username = local.public_pgsql_admin_login
+  password = random_password.pgsql_admin_password.result
```
timja, line 10: is there a chicken and egg problem here?

can you initialise the provider if the password doesn't exist yet?

dduportal: Good question! There is no egg and chicken problem at all since Terraform manages its tree of resources, and providers are nodes in the tree in the same way as the resource objects (or data objects).

Then, it's up to the provider implementation to support, or not, being put in line when Terraform walks the objects tree.
There is a nice explanation here: https://github.com/hashicorp/terraform/issues/2430#issuecomment-1054487362


In the case here, since we already created the postgres server, we cannot be 100% sure.

However:
- The test on the staging environment (which is always a "terraform plan" against an empty environment) makes us safe neough to say that the "dry run" implementation of the providers is perfectly fine (itherwise the test of the "terraform plan" would fail)
- The provider maintainer should have resolved this starting with the 1.10 version (currently 1.15): https://github.com/cyrilgdn/terraform-provider-postgresql/issues/103#issuecomment-880647860
=> I'm pretty confident we are covered. Worst case we would have to create in 2 steps next time we start from scratch ;)
-

timja (approved):

PR #181, opened 2022-03-28
add the output for dbconfig secrets

smerle33:
Part of https://github.com/jenkins-infra/helpdesk/issues/1627 (migrating ratings.jenkins.io from AWS VM to Azure AKS).

terraform output the dbconfig.php


rating.jenkins.io.tf
```diff
@@ -12,3 +12,16 @@ resource "postgresql_role" "rating" {
...
+  description = "file /config/dbconfig.php"
+  value       = <<-EOT
+    <?php
+      $dbuser='${local.public_pgsql_admin_login}';
+      $dbpass='${random_password.pgsql_rating_user_password.result}';
+      $dbname='rating';
+      $dbserver='${azurerm_postgresql_flexible_server.public.fqdn}';
+      $dbport='';
+  EOT
+}
```
lemeurherve, line 27:
```suggestion
}

```
(Ending new line)

lemeurherve (dismissed):

lemeurherve (approved):
I would have skipped the final PHP tag, but if it's works like that go for it

timja (commented):
What is this doing? It doesn’t seem to be used?

smerle33:
> What is this doing? It doesn’t seem to be used?

can be avoid: https://www.php.net/manual/en/language.basic-syntax.instruction-separation.php

lemeurherve:
> I mean what is this output for? how is it going to be consumed?

By hand ^^

Ideally, it would be registered as secret by this pipeline, but it implies giving access to sops, which is too risky.
So instead this output will be manually saved in jenkins-infra/charts-secrets to be consumed later by the rating helm chart.

In the long term it would be nice to use a vault.

lemeurherve (approved):

timja:
> By hand ^^

this seems unneeded then?

at least comment why it's there?

dduportal:
> > By hand ^^
> 
> this seems unneeded then?
> 
> at least comment why it's there?

By hand meaning "copy and paste the output to sops when the user or database is re-created", as this will be a multi-line secret to be encrypted (and then mounted as secret in the upcoing rating helm chart).

But good point about the comment: ping @smerle33 can you add a comment right before the "output" block explaining it destination please?

rating.jenkins.io.tf
```diff
@@ -12,3 +12,16 @@ resource "postgresql_role" "rating" {
   login    = true
   password = random_password.pgsql_rating_user_password.result
 }
+
+output "dbconfig" {
```
dduportal, line 16:
```suggestion
# This (sensitive) output is meants to be encrypted into the production secret system, to be provided as a secret to the ratings.jenkins.io application
output "rating_dbconfig" {
```

dduportal: Suggestion to add:
- A comment (as per @timja comment) explaining the role of the output
- rename the output by adding "rating" as prefix, since the output namespace is global to the whole repository

timja (approved):

rating.jenkins.io.tf
```diff
@@ -12,3 +12,17 @@ resource "postgresql_role" "rating" {
   login    = true
   password = random_password.pgsql_rating_user_password.result
 }
+
+# This (sensitive) output is meants to be encrypted into the production secret system, to be provided as a secret to the ratings.jenkins.io application
```
lemeurherve, line 16:
```suggestion
# This sensitive output is meant to be encrypted into the production secret system, to be provided as a secret to the ratings.jenkins.io application
```

lemeurherve (approved):

PR #183, opened 2022-03-28
Fix/public pgsql net

smerle33:
provide a private dns zone and a link between virtual network and private dns zone.


dduportal (approved):
LGTM!

PR #218, opened 2022-11-17
breaking(privatek8s) apply Azure Best Practises on node pools (ephemeral system disks and avoid multi-AZ due to PVs)

dduportal:
(edit) 💥 this PR destroyes and create a new cluster + IP address (ref. https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#os_disk_type => `Changing this forces a new resource to be created`)

As per @lemeurherve diagnostics on the newly created cluster (`privatek8s`), the node pools should use ephemeral system disks as indicated by Azure Best Practises:

image

This PR sets the system disk type to Ephemeral instead of the default "Managed" as per https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster_node_pool#os_disk_type.

(edit 2)
Since this PR will re-create the cluster, I've also added another change: we only deploy the node pools to the same zone `3`:

image

- We do not need "multi-zone" for the private k8s (infra.ci/releasE.ci). In case of a zone being  we accept this cluster to be inoperative until fixed (unless we create ZRS PVC as per https://learn.microsoft.com/en-us/azure/aks/availability-zones#azure-disk-availability-zone-support)

### DIFF
--- a/privatek8s.tf
+++ b/privatek8s.tf
@@ -38,6 +38,7 @@ resource "azurerm_kubernetes_cluster" "privatek8s" {
   default_node_pool {
     name           = "systempool"
     vm_size        = "Standard_D2as_v4"
+    os_disk_type   = "Ephemeral"
     node_count     = 1
     vnet_subnet_id = azurerm_subnet.privatek8s_tier.id
     tags           = local.default_tags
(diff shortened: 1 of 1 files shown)


privatek8s.tf
```diff
@@ -38,9 +38,11 @@ resource "azurerm_kubernetes_cluster" "privatek8s" {
   default_node_pool {
     name           = "systempool"
     vm_size        = "Standard_D2as_v4"
+    os_disk_type   = "Ephemeral"
     node_count     = 1
     vnet_subnet_id = azurerm_subnet.privatek8s_tier.id
     tags           = local.default_tags
+    zones          = [3]
```
lemeurherve, line 45: Out of curiosity, why the third zone and not one of the others?

lemeurherve (approved):
❤️ 

(I've tainted the random pet so we'll get a new cluster name)

PR #238, opened 2022-12-01
feat: add `updatecli` user-assigned managed identity

lemeurherve:
https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview

### DIFF
--- a/managed-identities.tf
+++ b/managed-identities.tf
@@ -1,0 +1,12 @@
+resource "azurerm_resource_group" "accounts" {
+  name     = "managed-identities"
+  location = var.location
+  tags = local.default_tags
+}
+
+# To be used by updatecli for listing resources like OS image versions for example
+resource "azurerm_user_assigned_identity" "updatecli" {
+  name                = "updatecli"
+  location            = azurerm_resource_group.accounts.location
+  resource_group_name = azurerm_resource_group.accounts.name
+}


timja (commented):
where is this going to be used? on an azure vm? or kubernetes pod? or somewhere else?

lemeurherve:
I'd like to use it in a kubernetes pod to list available OS versions so updatecli could update the VPN VM: https://github.com/jenkins-infra/azure-net/blob/3f3d5f15e1f8534b94e56b2e16db4103031d8854/vpn.tf#L158-L159

I need to execute `az login` before being able to list them, so I though about using a managed identity to avoid interacting  directly with a service principal like we're currently doing for cleaning packer images: https://github.com/jenkins-infra/packer-images/blob/61405aca644e90ab22c8643114b13ec79c8bd888/Jenkinsfile_k8s#L99-L115

Is there a better way or does it looks about right to you?

timja:
Makes sense although you probably want to add a federated credential:
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/federated_identity_credential

Then use Azure AD workload identity so the pod can authenticate with the identity:
https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview

timja:
Also will need to enable `workload_identity_enabled` and `oidc_issuer_enabled` on the cluster:


https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#workload_identity_enabled

dduportal:
Do we agree that it would require either:
- To define an svcaccount (mapping to the workload identity) only for the pods running this updatecli task (or define multiple namespaces for agents)
- Or to accept that all of the pods (agents) are using this identity
?

Side note @lemeurherve please check that this change won't delete/create the cluster. With EKS it did so better safe than sorry :)

timja:
> Side note @lemeurherve please check that this change won't delete/create the cluster. With EKS it did so better safe than sorry :)

We enabled it online on our clusters

----

> To define an svcaccount (mapping to the workload identity) only for the pods running this updatecli task (or define multiple namespaces for agents)

I would assume this ^ if it's not too much overhead.

dduportal:
This issue is old and does not map to a clearly defined use case. I'm closing it.

PR #240, opened 2022-12-02
feat: add an archive storage account

lemeurherve:


archives.tf
```diff
@@ -0,0 +1,22 @@
...
+  allow_nested_items_to_be_public = false
+  public_network_access_enabled   = false
+  # https://learn.microsoft.com/en-gb/azure/storage/common/infrastructure-encryption-enable
+  infrastructure_encryption_enabled = true
+
+  tags = local.default_tags
+}
+
+# Dump of confluence databases, see https://github.com/jenkins-infra/helpdesk/issues/3249
+# Should containers be created as code or manually?
```
timja, line 22: why would you not create as code?

dduportal: +1 with creating containetr as code. There is no sensitive information in the container definition itself.

dduportal (dismissed):
👍

dduportal:
@lemeurherve can you remove the changes on the terraform lock file? It only adds the checksum of the macOS Intel binaries for providers which ill be overwritten later (and it creates conflicts)

PR #245, opened 2022-12-07
fix: restrict access to `archives` storage account

lemeurherve:
can't access containers otherwise


dduportal (changes requested):
Not sure we want this. We want items to be private, which would require setting carefully the permissions

timja (dismissed):
I assume this is temporary until the TODO is solved:

> TODO: set to false and add a network rule allowing only the VPN as soon as this one has been deployed so we can access containers in the mean time.

lemeurherve:
@timja indeed. I went ahead and added network rules, allowing only a set of our IPs and the subnet where temp-privatek8s is located.

Note: I had to add a sevice endpoint on this subnet virtual network to avoid a NetworkAclsValidationFailure:

image

dduportal (approved):
`No changes. Your infrastructure matches the configuration.` => LGTM

PR #257, opened 2023-01-19
feat(privatek8s): enable KeyVault service endpoint

lemeurherve:
Enabling this endpoint so the privatek8s subnet can be added to the prodjenkinsinfra keys vault allowed networks, in order for infra.ci.jenkins.io to access secrets to control and admin other clusters.

Ref: https://github.com/jenkins-infra/helpdesk/issues/2844

### DIFF
--- a/privatek8s.tf
+++ b/privatek8s.tf
@@ -18,6 +18,8 @@ data "azurerm_subnet" "privatek8s_tier" {
   name                 = "privatek8s-tier"
   resource_group_name  = data.azurerm_resource_group.private.name
   virtual_network_name = data.azurerm_virtual_network.private.name
+  # Enable KeyVault service endpoint so the cluster can access secrets to update other clusters
+  service_endpoints = ["Microsoft.KeyVault"]
 }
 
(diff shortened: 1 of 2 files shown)


privatek8s.tf
```diff
@@ -18,6 +18,8 @@ data "azurerm_subnet" "privatek8s_tier" {
   name                 = "privatek8s-tier"
   resource_group_name  = data.azurerm_resource_group.private.name
   virtual_network_name = data.azurerm_virtual_network.private.name
+  # Enable KeyVault service endpoint so the cluster can access secrets to update other clusters
+  service_endpoints = ["Microsoft.KeyVault"]
```
timja, line 22: this is a data resource...? you can't do that here

lemeurherve:
Wrong repo, should be done on the resource in azure-net, not the data here.

privatek8s.tf
lemeurherve: https://github.com/jenkins-infra/azure-net/pull/34

PR #282, opened 2023-03-06
chore(publick8s): ensure that storage class allow volume expansion is explicitly defined

smerle33:
https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/storage_class#allow_volume_expansion

as per https://github.com/jenkins-infra/helpdesk/issues/3425#issuecomment-1456373704

### DIFF
--- a/publick8s.tf
+++ b/publick8s.tf
@@ -107,6 +107,7 @@ resource "kubernetes_storage_class" "managed_csi_premium_retain_public" {
     skuname = "Premium_LRS"
   }
   provider = kubernetes.publick8s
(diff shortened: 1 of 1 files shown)


lemeurherve (commented):
This is the default value of Azure Storage Class 🤷

dduportal (approved):
LGTM!

Should not change anything (as per the terraform plan)

dduportal:
> This is the default value of Azure Storage Class 🤷

Yes, but it is ALSO the default on AWS and it was not enabled until mad explicit...

PR #285, opened 2023-03-08
fix(trusted.ci) Ensure that "terraform-production" is the explicit owner of the azuread_application

dduportal:
This PR is assorted of the following changes:

- State was cleaned to remove `azuread_application.packer`
- The trusted.ci.jenkins.io SP was re-imported in the state

### DIFF
--- a/versions.tf
+++ b/versions.tf
@@ -5,6 +5,9 @@ terraform {
     azurerm = {
       source = "hashicorp/azurerm"
     }
+    azuread = {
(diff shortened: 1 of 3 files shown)


versions.tf
```diff
@@ -17,5 +20,8 @@ terraform {
     random = {
       source = "hashicorp/random"
     }
+    time = {
```
hervelemeur, line 23: What is it for? Implied dependency?

PR #292, opened 2023-03-15
feat(privatek8s): add `azurefile-csi-premium-retain` storage class

lemeurherve:
Same mount options as https://github.com/jenkins-infra/helm-charts/blob/a491e7356a304fc2545b175dc90b880a50b727fc/charts/env-jenkins-release/templates/core-packages.yaml#L29-L35, will be used to set storage class in this chart for `release` volumes.

Ref: https://github.com/jenkins-infra/helpdesk/issues/2844


dduportal (approved):
🚀

PR #317, opened 2023-03-30
feat(trusted): restricting SP score to his resourceGroup

smerle33:


dduportal (changes requested):
The current change (restricitng the scope to resource group, for the "Contribuor" role) is ✅ 

However, this PR must also add the permission to read the VM images from the packer resource group, otherwise merging this will prevent trusted.ci.jenkins.io from creating Azure VM agents.

Ref. https://github.com/jenkins-infra/azure/blob/2454778c58a806dc4873ea891b29609df3d6e39e/cert.ci.jenkins.io.tf#L62-L66

dduportal (approved):
`Plan: 2 to add, 0 to change, 1 to destroy.` 


LGTM :+1:

Dion't forget to check trusted.ci after applying + a controller restart from the VM to ensure a new Azure session is regenerated (using the SP credential which did not change)

PR #326, opened 2023-04-13
feat(shared images galleries): add amd64 and arm64 galleries

smerle33:
add galleries for amd64 and arm64 for windows also as it's now mandatory to have the architecture within the gallery name for our packer to upload images.
windows ARM64 is not used yet.

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -13,7 +13,11 @@ locals {
         "ubuntu-22.04-amd64" = "eastus"
         "ubuntu-22.04-arm64" = "eastus"
         "windows-2019"       = "eastus"
+        "windows-2019-amd64" = "eastus"
+        "windows-2019-arm64" = "eastus"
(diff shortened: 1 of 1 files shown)


timja (approved):

dduportal (commented):
To be verified, but I don't believe there are Windows Server ARM64 images.

smerle33:
> To be verified, but I don't believe there are Windows Server ARM64 images.

I was more thinking on more folders in case we will use them WDYT ? ... I also changed all region to us-east 

Plan: 13 to add, 0 to change, 1 to destroy.

dduportal:
> I was more thinking on more folders in case we will use them WDYT?

We don't need them, so better not creating them empty. If we need them, it'll be easy to add them

smerle33:
> > I also changed all region to us-east
> 
> What is the impact ?

FWIT none ...

PR #329, opened 2023-04-21
fix: increase linuxpool disk size from 30GB to 40GB

lemeurherve:
Following the switch to a percentage based alerting for the available disk size in https://github.com/jenkins-infra/datadog/pull/168, we've noticed the available disk space on linuxpool is below 20% (currently at 16,7%)

This PR is increasing the OS disk size on this node pool.

Ref: https://github.com/jenkins-infra/helpdesk/issues/3497

PR #330, opened 2023-04-22
 fix(privatek8s) increase linuxpool disk size from 40GB to 50GB

dduportal:
Follow up of https://github.com/jenkins-infra/azure/pull/329: it had a positive impact but we still receive warnings.

### DIFF
--- a/privatek8s.tf
+++ b/privatek8s.tf
@@ -63,7 +63,7 @@ resource "azurerm_kubernetes_cluster_node_pool" "linuxpool" {
   name                  = "linuxpool"
   vm_size               = "Standard_D4s_v3"
   os_disk_type          = "Ephemeral"
-  os_disk_size_gb       = 40
+  os_disk_size_gb       = 50
   kubernetes_cluster_id = azurerm_kubernetes_cluster.privatek8s.id
   enable_auto_scaling   = true
   min_count             = 0


lemeurherve:
Heh, synchronized ^^
https://github.com/jenkins-infra/datadog/pull/172#issuecomment-1518594114

dduportal:
Ok, so this PR also broke the linuxpool, like #329 despite the fix you made in jenkins-infra/pipeline-library.

Currently fixing (same method as last time: terraform apply from local machine)

dduportal:
- First tentative:
  - Remove the `lock` with the command
  - Applied locally: the nodepool was recreated
  - After ~5-6 minte, infra.ci was back. Re-triggered the azure terraform job which failed with the same behavior (deprovisionning the linuxpool).
- Checked the workload: terraform agents were strill running on the linuxpool: make sense since there a no taints on this node pool, and the toleration added in https://github.com/jenkins-infra/pipeline-library/pull/655 are not sufficient. I did not realize when reviewing and merging it, so no worries, let's learn:
  - Toleration allows to run on some nodes which have tains IF the scheduler select these nodes. But it does not mean these nodes will be selected (other can).
- Added https://github.com/jenkins-infra/pipeline-library/commit/999a2f18edd0328f92727b0cb2b627c05b5035e5 as hotfix to add a nodeselector to "hint" the scheduler on which nodepool to schedule (along with toleration).

- Second retry in progress

dduportal:
- infra.ci.jenkins.io is back 👍 
- 2 hotfixes were required to allow scheduling terraform pod agent in the `infracipool`:
  - Adding nodeselector: https://github.com/jenkins-infra/pipeline-library/commit/999a2f18edd0328f92727b0cb2b627c05b5035e5
  - Adding missing toleration for `spot`: https://github.com/jenkins-infra/pipeline-library/commit/1ac7686c4baa80799034b89da2375744c7a0363f
- Note: don't forget to clear cache of the `pipeline-library` in Manage Jenkins to see the hotfix changes 😅
- Note: the `updatecli` pods are not scheduled on the `infracipool`. I chose not to jump in this topic: we should switch to admin defined pods to define tolerations and affinities, and or update the shared libraries to pass these "environment related setup" as parameter.

Capture d’écran 2023-04-22 à 13 50 11


=> we should be able to validate that we can now safely operate the linuxpool node (without breaking the terraform process) once we'll upgrade the node pool OS image to Ubuntu 22.04 (https://github.com/jenkins-infra/helpdesk/issues/2982)

PR #334, opened 2023-04-27
feat(bounce/trusted): add network rules for bounce

smerle33:
this set some network rules to allow ssh from admin and deny everything else

### DIFF
--- a/trusted.ci.jenkins.io.tf
+++ b/trusted.ci.jenkins.io.tf
@@ -121,3 +121,54 @@ resource "azurerm_linux_virtual_machine" "trusted_bounce" {
     version   = "latest"
   }
 }
+
+####################################################################################
+## Network Security Groups for TRUSTED subnets
+####################################################################################
+# subnet trusted_controller
+resource "azurerm_network_security_group" "trusted_controller" {
+  name                = data.azurerm_subnet.trusted_controller.name
+  location            = data.azurerm_resource_group.trusted.location
+  resource_group_name = data.azurerm_resource_group.trusted.name
+
+  # No security rule: using 'azurerm_network_security_rule' to allow composition across files
+
+  tags = local.default_tags
+}
+
+resource "azurerm_subnet_network_security_group_association" "trusted_controller" {
+  subnet_id                 = data.azurerm_subnet.trusted_controller.id
+  network_security_group_id = azurerm_network_security_group.trusted_controller.id
+}
+
+resource "azurerm_network_security_rule" "deny_all_to_vnet" {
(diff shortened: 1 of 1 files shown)


trusted.ci.jenkins.io.tf
```diff
@@ -121,3 +121,54 @@ resource "azurerm_linux_virtual_machine" "trusted_bounce" {
...
+resource "azurerm_subnet_network_security_group_association" "trusted_controller" {
+  subnet_id                 = data.azurerm_subnet.trusted_controller.id
+  network_security_group_id = azurerm_network_security_group.trusted_controller.id
+}
+
+resource "azurerm_network_security_rule" "deny_all_to_vnet" {
+  name = "deny-all-to-vnet"
+  # Priority should be the highest value possible (lower than the default 65000 "default" rules not overidable) but higher than the other security rules
+  # ref. https://github.com/hashicorp/terraform-provider-azurerm/issues/11137 and https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/network_security_rule#priority
+  priority                     = 4095
```
lemeurherve, line 148: Out of curiosity, where this number comes from?

lemeurherve (approved):

dduportal (approved):
LGTM

trusted.ci.jenkins.io.tf
```diff
@@ -121,3 +121,54 @@ resource "azurerm_linux_virtual_machine" "trusted_bounce" {
...
+  source_address_prefix        = "*"
+  destination_address_prefixes = data.azurerm_virtual_network.trusted.address_space
+  resource_group_name          = data.azurerm_resource_group.trusted.name
+  network_security_group_name  = azurerm_network_security_group.trusted_controller.name
+}
+
+resource "azurerm_network_security_rule" "allow_ssh_from_admins_to_bounce" {
+  for_each = local.admin_allowed_ips
+
+  name                        = "allow-22-from-${each.key}-to-bounce"
```
timja, line 163: this is missing a number of people that likely use trusted.ci.

this is not something we have restricted on AWS why is it being added during a like for like migration?

dduportal: Fair point. Any objection to keep the restriction during the initial setup and remove it once it's working as expected?
(at the expanse of adding a TODO comment here of course)

smerle33: How, sorry not to have pointed that out, the goal was to lock for us 3 first during the preparation.
TODO: add all used ips or remove the rule at final migration.
Thanks @timja

PR #333, opened 2023-04-25
feat(trusted): add bounce vm on azure

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3486

create a bounce VM on azure 

from https://github.com/jenkins-infra/azure/pull/331

PR #331, opened 2023-04-24
feat(trusted): first step to deploy a trusted infra on azure

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3486
this PR is the first round to provide 3 VMs : 

- bounce VM (can be replace later by a vpn)
- controller VM
- permanent agent VM

PR #337, opened 2023-04-28
feat(trusted): controller VM

smerle33:
following https://github.com/jenkins-infra/azure/pull/333 and https://github.com/jenkins-infra/azure/pull/334

adding a controller VM

### DIFF
--- a/trusted.ci.jenkins.io.tf
+++ b/trusted.ci.jenkins.io.tf
@@ -137,7 +137,7 @@ resource "azurerm_network_interface" "trusted_controller" {
     subnet_id                     = data.azurerm_subnet.trusted_controller.id
     private_ip_address_allocation = "Static"
     # TODO add a dns record to point on this private ip
-    private_ip_address            = "10.252.0.1" # Manually chosen first IP of the controller subnet
+    private_ip_address = "10.252.0.1" # Manually chosen first IP of the controller subnet
   }
 }
 
--- a/trusted.ci.jenkins.io.tf
+++ b/trusted.ci.jenkins.io.tf
@@ -69,6 +69,8 @@ resource "azurerm_role_assignment" "trusted_ci_jenkins_io_allow_packer" {
   principal_id         = azuread_service_principal.trusted_ci_jenkins_io.id
 }
 
+# VMs
+# BOUNCE VM
 resource "azurerm_public_ip" "trusted_bounce" {
   name                = "trusted-bounce"
   location            = azurerm_resource_group.trusted_ci_jenkins_io_controller.location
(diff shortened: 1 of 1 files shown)


lemeurherve (approved):

trusted.ci.jenkins.io.tf
```diff
@@ -122,6 +124,72 @@ resource "azurerm_linux_virtual_machine" "trusted_bounce" {
...
+resource "azurerm_network_interface" "trusted_controller" {
+  name                = "trusted-controller"
+  location            = azurerm_resource_group.trusted_ci_jenkins_io_controller.location
+  resource_group_name = azurerm_resource_group.trusted_ci_jenkins_io_controller.name
+  tags                = local.default_tags
+
+  ip_configuration {
+    name                          = "internal"
+    subnet_id                     = data.azurerm_subnet.trusted_controller.id
+    private_ip_address_allocation = "Dynamic"
```
dduportal, line 138: I wonder if we should not select "static" and define a static IP in the subnet, as it is the case today: https://github.com/jenkins-infra/runbooks/tree/main/trusted-ci#connecting 
(otherwise the ssh config of all users would have to change on each VM restart where the private IP could change).
Unless I misunderstood the lifecycle of this ip dynamic allocation (maybe it is only if we chaneg the NIC)?

A safer road would be to define an (internal) DNS record in this subnet (e.g. only inside this private vnet) pointing to the private ip of the VM: in that case no problem to have it dynamically defined, as soon as we let the DNS record to always match the IP or the VM.

WDYT?

(Non blocking comment: we can start with the actual code)

trusted.ci.jenkins.io.tf
```diff
@@ -122,6 +124,72 @@ resource "azurerm_linux_virtual_machine" "trusted_bounce" {
...
+  ]
+
+  admin_ssh_key {
+    username   = local.trusted_ci_jenkins_io.admin_username
+    public_key = "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQC5K7Ro7jBl5Kc68RdzG6EXHstIBFSxO5Da8SQJSMeCbb4cHTYuBBH8jNsAFcnkN64kEu+YhmlxaWEVEIrPgfGfs13ZL7v9p+Nt76tsz6gnVdAy2zCz607pAWe7p4bBn6T9zdZcBSnvjawO+8t/5ue4ngcfAjanN5OsOgLeD6yqVyP8YTERjW78jvp2TFrIYmgWMI5ES1ln32PQmRZwc1eAOsyGJW/YIBdOxaSkZ41qUvb9b3dCorGuCovpSK2EeNphjLPpVX/NRpVY4YlDqAcTCdLdDrEeVqkiA/VDCYNhudZTDa8f1iHwBE/GEtlKmoO6dxJ5LAkRk3RIVHYrmI6XXSw5l0tHhW5D12MNwzUfDxQEzBpGK5iSfOBt5zJ5OiI9ftnsq/GV7vCXfvMVGDLUC551P5/s/wM70QmHwhlGQNLNeJxRTvd6tL11bof3K+29ivFYUmpU17iVxYOWhkNY86WyngHU6Ux0zaczF3H6H0tpg1Ca/cFO428AVPw/RTJpcAe6OVKq5zwARNApQ/p6fJKUAdXap+PpQGZlQhPLkUbwtFXGTrpX9ePTcdzryCYjgrZouvy4ZMzruJiIbFUH8mRY3xVREVaIsJakruvgw3b14oQgcB4BwYVBBqi62xIvbRzAv7Su9t2jK6OR2z3sM/hLJRqIJ5oILMORa7XqrQ=="
+  }
+
+  os_disk {
+    caching              = "ReadWrite"
+    storage_account_type = "Standard_LRS"
```
dduportal, line 162: Controller machine requires an SSD (different than bounce) due to the required latency constraints for the system

trusted.ci.jenkins.io.tf
```diff
@@ -122,6 +124,72 @@ resource "azurerm_linux_virtual_machine" "trusted_bounce" {
...
+    sku       = "minimal-22_04-lts-gen2"
+    version   = "latest"
+  }
+}
+
+resource "azurerm_managed_disk" "trusted_controller_data_disk" {
+  name                 = "trusted-controller-data-disk"
+  location             = azurerm_resource_group.trusted_ci_jenkins_io_controller.location
+  resource_group_name  = azurerm_resource_group.trusted_ci_jenkins_io_controller.name
+  storage_account_type = "Standard_LRS"
```
dduportal, line 178: SSD required for the datadisk hosting Jenkins Home

dduportal (changes requested):
A few comments, only blockers are the SSDs. The rest of the PR code looks really good at first glance!

trusted.ci.jenkins.io.tf
```diff
@@ -122,6 +124,74 @@ resource "azurerm_linux_virtual_machine" "trusted_bounce" {
   }
 }
 
+# CONTROLLER VM
+## NETWORK INTERFACE with internal ip
+resource "azurerm_network_interface" "trusted_controller" {
+  name                = "trusted-controller"
```
dduportal, line 130: Not blocking this PR, but need to be addressed before migration to this new machine:

The naming `trusted-*` for these VMs can be confusing:

- Is `trusted` an adjective or a name?
- How should we name `ci.jenkins.io` VM? `ci-controller`? Then it's not coherent with the naming here.

Proposal: ensure that the `ci` string is included, but do not add the `jenkins.io` domain as it is subject to change (and add too much elements in the name). WDYT?

- `trusted-ci-controller` (Terraform resource name: `trusted_ci_controller`)
- For ci.jenkins.io: `ci-controller` (`ci_controller`)
- For cert.ci.jenkins.io): `cert-ci-controller` (`trusted_ci_jenkins_io_controller`)
- Bounce would be named to `trusted-ci-bounce` (I would even argue for `trusted-ci-bastion`)

trusted.ci.jenkins.io.tf
```diff
@@ -122,6 +124,74 @@ resource "azurerm_linux_virtual_machine" "trusted_bounce" {
...
+  }
+}
+
+resource "azurerm_managed_disk" "trusted_controller_data_disk" {
+  name                 = "trusted-controller-data-disk"
+  location             = azurerm_resource_group.trusted_ci_jenkins_io_controller.location
+  resource_group_name  = azurerm_resource_group.trusted_ci_jenkins_io_controller.name
+  storage_account_type = "StandardSSD_LRS"
+  create_option        = "Empty"
+  disk_size_gb         = "100"
```
dduportal, line 182:
```suggestion
  disk_size_gb         = "128"
```

As per https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types#standard-ssd-size (better use as much data as what we pay for)

dduportal (approved):

trusted.ci.jenkins.io.tf
timja: https://github.com/jenkins-infra/azure/pull/331#pullrequestreview-1398165722 would make this a lot simpler and less to review imo

dduportal: @timja I'm having mixed feelings about a module as the 3 controllers have 3 very different infrasturcture topologies: 1 public, 1 private with SSH bounce and 1 private behind a VPN.

Not sure about the value of "DRY" as most of the lines have different values :|

dduportal: (btw sorry we missed your comment on the closed PR: do you mind commenting on the issue to have the discussion in a place were it would be easier to keep and track it?)

dduportal: (Don't get me wrong, I don't see a module would not be usefull, I'm questionning the value "now" for the trusted migration: I would have that discussion with https://github.com/jenkins-infra/helpdesk/issues/3535 for instance, comparing if it is worth it with the trusted machine)

dduportal (approved):
LGTM 👏

PR #338, opened 2023-05-02
feat(trusted): add controller VM with dynamic ip with dynamic DNS record for this private ip

smerle33:

### DIFF
--- a/trusted.ci.jenkins.io.tf
+++ b/trusted.ci.jenkins.io.tf
@@ -138,14 +138,6 @@ resource "azurerm_network_interface" "trusted_ci_controller" {
     private_ip_address_allocation = "Dynamic"
   }
 }
-resource "azurerm_dns_a_record" "trusted_ci_controller" {
-  name                = "trusted-ci-controller"
-  zone_name           = data.azurerm_dns_zone.jenkinsio.name
-  resource_group_name = data.azurerm_resource_group.proddns_jenkinsio.name
-  ttl                 = 300
-  records             = [azurerm_network_interface.trusted_ci_controller.private_ip_address]
-  tags                = local.default_tags
-}
 
 ## MACHINE (controller)
 resource "azurerm_linux_virtual_machine" "trusted_ci_controller" {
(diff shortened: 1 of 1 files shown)


trusted.ci.jenkins.io.tf
```diff
@@ -135,11 +135,17 @@ resource "azurerm_network_interface" "trusted_ci_controller" {
...
   }
 }
+resource "azurerm_dns_a_record" "trusted_ci_controller" {
+  name                = "trusted-ci-controller"
+  zone_name           = data.azurerm_dns_zone.jenkinsio.name
+  resource_group_name = data.azurerm_resource_group.proddns_jenkinsio.name
+  ttl                 = 300
+  records             = [azurerm_network_interface.trusted_ci_controller.private_ip_address]
+  tags                = local.default_tags
+}
```
dduportal, line 148:
```suggestion
```

Proposal:

- Remove this code block so we can merge a first version with the dyanmicaly allocated IP to unblock the main branch
- Open a subsequent PR with the DNS record that will need to be private. Hopefully, Azure provides the concept of "Private DNS zone" and "private records" that can be bound to virtual networks \o/ And we already have this kind of setup for pgsql (in `pgsql.tf`):
  - https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/private_dns_zone
  - https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/private_dns_a_record
  - https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/private_dns_zone_virtual_network_link 
  - https://github.com/jenkins-infra/azure/blob/011e04d35d1f43c7c8e9e812cd0083ca491d5759/pgsql.tf#L37-L47
  - The idea would be to have a private zone named `trusted.ci.jenkins.io` with the apex record set to the controller IP (and in the future, adding `A` record for the future permaneant agent such as `agent1.trusted.ci.jenkins.io`)

dduportal (changes requested):

trusted.ci.jenkins.io.tf
timja: Yes it does support private dns zones and linking them to specific vnets

dduportal (approved):

dduportal:
Merging to see a quick feedback (after infra.ci was broken, checks might be in poor state)

PR #339, opened 2023-05-04
feat(trusted): add a private dns record for the controller

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3486
and https://github.com/jenkins-infra/azure/pull/338#discussion_r1182794101
we defined a specific internal dns a record for the controller 

to be merge only when main will build

PR #341, opened 2023-05-09
chore(trusted): rename azurerm_private_dns_zone_virtual_network_link vnet

smerle33:
as per https://github.com/jenkins-infra/azure/pull/339#discussion_r1184965612 renaming to a more consistent version


timja (approved):

dduportal (approved):
=> the check failed for what seems to be a transient error due to a controller restart


LGTM with the hypothesis that a build replay would be green

PR #343, opened 2023-05-10
feat(trusted): security rules to allow internal SSH connections

smerle33:


dduportal (approved):
LGTM

A note (non blocking):

- WDYT about updating the title to replace `security rules` by `allow internal SSH connections` or something liek that (to make it easier to read in the future commit history)

dduportal (approved):

PR #345, opened 2023-05-10
feat(publick8s,privatek8s) use the maximum OS disk size available for user node pools

dduportal:


publick8s.tf
```diff
@@ -71,7 +71,7 @@ resource "azurerm_kubernetes_cluster_node_pool" "publicpool" {
   name                  = "publicpool"
   vm_size               = "Standard_D8s_v3" # 8 vCPU, 32 GB RAM, 64 GB disk, 16 000 IOPS
   os_disk_type          = "Ephemeral"
-  os_disk_size_gb       = 100
+  os_disk_size_gb       = 100 # Ref. Cache storage size at https://learn.microsoft.com/en-us/azure/virtual-machines/dv3-dsv3-series#dsv3-series (depends on the instance size)
```
lemeurherve, line 74:
```suggestion
  os_disk_size_gb       = 200 # Ref. Cache storage size at https://learn.microsoft.com/en-us/azure/virtual-machines/dv3-dsv3-series#dsv3-series (depends on the instance size)
```

lemeurherve: LGTM otherwise

lemeurherve (approved):

PR #348, opened 2023-05-15
feat(ci.jenkins.io) define a new VM + datadisk + public IP in a custom resource group

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3535, this PR is the first major step to create a new (faster + cheaper) ci.jenkins.io VM managed as code.

Please note the following elements:

- A new resource group is created to avoid messing up with the current one (to avoid any confusion). The goal will be to remove the older resource group once the migration will be complete
  - Neither DNS record, subnets or snapshot (of the data disk for migration) are constrained by this
- StandardSSD are used instead of PremiumSSD (V1/V2). The goal is to start with a cheaper storage and see the current IOPS usage (since we [removed the plugin-config-history plugin](https://github.com/jenkins-infra/helpdesk/issues/3528) and [added an S3 artifact caching proxy](https://github.com/jenkins-infra/helpdesk/issues/3496) the need for IOPS is lowered).
  - Keep using a 512 Gb SSD though, because current usage is ~ 240 Gb: if we want to stay under 80% usage, AND since [Azure managed disks](https://learn.microsoft.com/en-us/azure/virtual-machines/disks-types#standard-ssds) are either making you pay for 256 or 512 Gb (unless Premium SSD v2), better to have the bigger disk as possible.

- No need for a storage account (it was only used for boot diagnostics, which we should not need with Terrafomr IAC based VM)

- Registration to Puppet is done under the name `ci.jenkins.io` : the current VM is [registered as `azure.ci.jenkins.io`](https://github.com/jenkins-infra/jenkins-infra/blob/e425203e0ffafdcf8b2d5e675ad838c5e73cd687/manifests/site.pp#L76-L79) so no risk of conflicts

Please note that no security group is added...yet. I want to start with an "open VM" before applying security groups in a subsequent PR (everything will be removed and re-created once verified functionnal).

### DIFF
--- a/ci.jenkins.io.tf
+++ b/ci.jenkins.io.tf
@@ -2,6 +2,8 @@
 resource "azurerm_resource_group" "ci_jenkins_io_controller" {
   name     = "ci-jenkins-io-controller"
   location = "East US 2"
+  
+  tags = local.default_tags
 }
 
 resource "azurerm_resource_group" "ci_jenkins_io_agents" {
--- a/ci.jenkins.io.tf
+++ b/ci.jenkins.io.tf
@@ -73,7 +73,7 @@ resource "azurerm_linux_virtual_machine" "ci_jenkins_io" {
   os_disk {
     caching              = "ReadWrite"
     storage_account_type = "StandardSSD_LRS"
-    disk_size_gb         = 32 # Minimal size for ubuntu 22.04 image
+    disk_size_gb         = 64 # Minimal size for ubuntu 22.04 image
   }
 
   source_image_reference {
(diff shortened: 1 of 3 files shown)


ci.jenkins.io.tf
```diff
@@ -1,22 +1,94 @@
 /** Two resources groups: one for the controller, the second for the agents **/
-# resource "azurerm_resource_group" "ci_jenkins_io_controller" {
-#   name     = "prodjenkinsci"
-#   location = "East US 2"
-# }
+resource "azurerm_resource_group" "ci_jenkins_io_controller" {
+  name     = "ci-jenkins-io-controller"
+  location = "East US 2"
```
timja, line 4:
```suggestion
  location = "East US 2"
  
  tags = local.default_tags
```

timja (dismissed):

timja (approved):

PR #351, opened 2023-05-17
fix(trusted.ci.jenkins.io) add missing outbound rules to allow SSH internally

dduportal:
This PR is a fixup of #343 .

It allows SSH connections internally (bounce -> controller, controller ->.*-agents) to work as expected by opening the missing outbound connections in the security groups (the inbounds were already specified).

Please note that it also add a 2nd commit which rename the existing inbound rules to make it easier to visually differentiate inbound/outbound rules.

### DIFF
--- a/trusted.ci.jenkins.io.tf
+++ b/trusted.ci.jenkins.io.tf
@@ -399,10 +399,10 @@ resource "azurerm_network_security_rule" "deny_all_to_vnet" {
 }
 
 # TODO: add all ips from people needing access to trusted.ci.jenkins.io, or remove the rule at final migration. (Cf https://github.com/jenkins-infra/azure/pull/334#discussion_r1179955821)
-resource "azurerm_network_security_rule" "allow_ssh_from_admins_to_bounce" {
+resource "azurerm_network_security_rule" "allow_inbound_ssh_from_admins_to_bounce" {
   for_each = local.admin_allowed_ips
 
-  name                        = "allow-22-from-${each.key}-to-bounce"
+  name                        = "allow-inbound-ssh-from-${each.key}-to-bounce"
   priority                    = 4000 + index(keys(local.admin_allowed_ips), each.key)
   direction                   = "Inbound"
   access                      = "Allow"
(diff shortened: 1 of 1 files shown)


trusted.ci.jenkins.io.tf
lemeurherve, line 1: Which one of these 3 priorities should be "the highest" possible?

image

trusted.ci.jenkins.io.tf
```diff
@@ -334,6 +334,54 @@ resource "azurerm_subnet_network_security_group_association" "trusted_ci_control
   network_security_group_id = azurerm_network_security_group.trusted_ci_controller.id
 }
 
+resource "azurerm_network_security_rule" "allow_outbound_from_bounce_to_controller" {
+  name = "allow-outbound-from-bounce-to-controller"
```
lemeurherve, line 338:
```suggestion
resource "azurerm_network_security_rule" "allow_outbound_ssh_from_bounce_to_controller" {
  name = "allow-outbound-ssh-from-bounce-to-controller"
```

trusted.ci.jenkins.io.tf
```diff
@@ -351,10 +399,10 @@ resource "azurerm_network_security_rule" "deny_all_to_vnet" {
 }
 
 # TODO: add all ips from people needing access to trusted.ci.jenkins.io, or remove the rule at final migration. (Cf https://github.com/jenkins-infra/azure/pull/334#discussion_r1179955821)
-resource "azurerm_network_security_rule" "allow_ssh_from_admins_to_bounce" {
+resource "azurerm_network_security_rule" "allow_inbound_ssh_from_admins_to_bounce" {
   for_each = local.admin_allowed_ips
 
-  name                        = "allow-22-from-${each.key}-to-bounce"
+  name                        = "allow-inbound-ssh-from-${each.key}-to-bounce"
```
lemeurherve, line 405:
```suggestion
  name                        = "allow-inbound-ssh-from-admin-${each.key}-to-bounce"
```
Suggestion, not mandatory at all

trusted.ci.jenkins.io.tf
dduportal: I don't understand your comment? The 3 new outbound already have distinct priorities?

lemeurherve: It seems to me contradictory that each of the 3 comments L339, L355 & L371 state they should be the highest, and higher than the others.

timja (dismissed):

trusted.ci.jenkins.io.tf
dduportal: no problem, I wasn't sure to understand if it was about the comment, the priority values or both.

With your explanation, it's clear and it make sense.

As I was confused by these comments, and as these comments are a duplication of the documentation (both Azure and terraform azure), I choose to remove them and applying proper formatting.

dduportal:
Merging as:
- I got an approval
- The checks are OK
- The comments are answered

trusted.ci.jenkins.io.tf
dduportal: @lemeurherve if it is ok for you with the last commit (and merged version), you can resolve the conversation.

If it is not, May I ask you to open either an issue, or a PR (with a proposal)?

PR #354, opened 2023-05-19
feat: allow connections between postgres and publick8s

lemeurherve:


dduportal:
Alas this won't help for the migration:

- The subnet `pgsql-tiers` in which the Postgres managed instance runs is in the old network (with the overlap IP issues)
- Adding a link to the new network has the same effect as creating a network peering: it propagates the IP overlap issue  (which we don't want)
- Changing the delegated subnet of a a felxible postges instances will destroy / create the instance as per https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/postgresql_flexible_server#delegated_subnet_id (so we can't move the instance to a new subnet)

=> We'll have to create a new `postgresql_flexible_server` instance in a new subnet in the (new) `public-vnet` and migrate databases one by one from the current instance to the brand new

lemeurherve:
Closed in favor of https://github.com/jenkins-infra/azure/pull/356

PR #356, opened 2023-05-19
feat: add a postgres server

lemeurherve:
This PR adds a new postgres server to avoid network overlap issues with the existing "public" one (see https://github.com/jenkins-infra/azure/pull/354#issuecomment-1554549215).

Needs https://github.com/jenkins-infra/azure-net/pull/78

Ref: https://github.com/jenkins-infra/helpdesk/issues/3351

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -1,6 +1,6 @@
 locals {
-  public_pgsql_admin_login = "psqladmin${random_password.pgsql_admin_login.result}"
-  data_pgsql_admin_login   = "psqladmin${random_password.data_pgsql_admin_login.result}"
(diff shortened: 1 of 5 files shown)


dduportal (approved):
`Plan: 8 to add, 0 to change, 0 to destroy.`

Lets go


No need to change the instance sizing or setup as the Azure advisor reports no health issue, constant 55% usage of memory and less than 15% average of CPU usage.

dduportal:
The deployment of this PR failed with:

```
Flexible Server Name: "public-data"): polling after Create: Code="IPV6AddressSpaceNotSupportedForVnet" Message="The virtual network public-vnet has IPv6 address space which is currently not supported. Please try with a different virtual network that has only IPv4 address space."
```

🥲

PR #374, opened 2023-05-31
fix(uplink) correct output for proper connection chain

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3351#issuecomment-1568847649

This PR fixes the output for `uplink` and the default pgsql role to ease the migration.


dduportal (commented):

PR #376, opened 2023-06-01
feat(puppet): add dns record A and cname for puppet.jenkins.io

smerle33:
add A and CNAME dns records as code.

### DIFF
--- a/puppet.jenkins.io.tf
+++ b/puppet.jenkins.io.tf
@@ -111,3 +111,21 @@ resource "azurerm_linux_virtual_machine" "puppet_jenkins_io" {
     version   = "latest"
   }
 }
+
+resource "azurerm_dns_a_record" "azure_puppet_jenkins_io" {
+  name                = "azure.puppet.jenkins.io."
+  zone_name           = data.azurerm_dns_zone.jenkinsio.name
+  resource_group_name = data.azurerm_resource_group.proddns_jenkinsio.name
+  ttl                 = 300
+  records             = [azurerm_public_ip.puppet_jenkins_io.ip_address]
+  tags                = local.default_tags
+}
+
+resource "azurerm_dns_cname_record" "jenkinsio_target_puppet_jenkins_io" {
+  name                = "puppet.jenkins.io"
+  zone_name           = data.azurerm_dns_zone.jenkinsio.name
+  resource_group_name = data.azurerm_resource_group.proddns_jenkinsio.name
+  ttl                 = 60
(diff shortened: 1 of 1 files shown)


puppet.jenkins.io.tf
```diff
@@ -113,7 +113,7 @@ resource "azurerm_linux_virtual_machine" "puppet_jenkins_io" {
 }
 
 resource "azurerm_dns_a_record" "azure_puppet_jenkins_io" {
-  name                = "azure.puppet.jenkins.io."
+  name                = "azure.puppet" #azure.puppet.jenkins.io --> A
```
lemeurherve, line 116:
```suggestion
  name                = "azure.puppet"
```
Implicit, redundant with the resource type

puppet.jenkins.io.tf
```diff
@@ -122,7 +122,7 @@ resource "azurerm_dns_a_record" "azure_puppet_jenkins_io" {
 }
 
 resource "azurerm_dns_cname_record" "jenkinsio_target_puppet_jenkins_io" {
-  name                = "puppet.jenkins.io"
+  name                = "puppet" #puppet.jenkins.io --> CNAME vers azure.puppet.jenkins.io
```
lemeurherve, line 125:
```suggestion
  name                = "puppet"
```
Same as above

puppet.jenkins.io.tf
```diff
@@ -111,3 +111,21 @@ resource "azurerm_linux_virtual_machine" "puppet_jenkins_io" {
...
+  ttl                 = 300
+  records             = [azurerm_public_ip.puppet_jenkins_io.ip_address]
+  tags                = local.default_tags
+}
+
+resource "azurerm_dns_cname_record" "jenkinsio_target_puppet_jenkins_io" {
+  name                = "puppet" #puppet.jenkins.io --> CNAME vers azure.puppet.jenkins.io
+  zone_name           = data.azurerm_dns_zone.jenkinsio.name
+  resource_group_name = data.azurerm_resource_group.proddns_jenkinsio.name
+  ttl                 = 60
```
lemeurherve, line 128: Why this TTL is set to 60 while the other to 300?
Because they are records of different types?

dduportal (approved):
LGTM!

I'm deleting the existing records (manually managed) as the service is already down

puppet.jenkins.io.tf
lemeurherve, line 128: Why two different TTLs?

dduportal: I guess it's a mistake, gotta fix it in #377 

Good catch

PR #380, opened 2023-06-02
feat(trusted.ci) add NAT gateway and DNS record to allow deterministic outbound

dduportal:
Part of https://github.com/jenkins-infra/helpdesk/issues/3486#issuecomment-1573305761

Despite the LDAP allowed IP being updated to the apparent outbound public IP ofr the controller.trusted VM, the connection is still not allowed.

As explained in https://learn.microsoft.com/en-us/azure/virtual-network/ip-services/default-outbound-access, using the default outbound IP (that is the case of the controller the agents as they do not have Public IPs) is not sustainable:

- The IP can change at any moment
- Microsoft could provide different outbound path with different apparent IPs (which seems to be our case)
- Scalibility is not there.

This PR adds a NAT gateway solution as it is recommended.

----

Notes:

- Our AKS clutsers are using a Loadbalancer outbound IP, which is their default. Based on what we'll learn for NAT gateway, we might need to do the same
- Determisitc outbound IPs is a hard requirement for us to understand how do we consume bandwidth in JFrog


timja (commented):
what does it connect to which needs a static outbound ip?

dduportal (commented):
`Plan: 7 to add, 0 to change, 0 to destroy.` looks good

dduportal:
> what does it connect to which needs a static outbound ip?

Sorry I've edited the PR body while you posted your message.

The answers are:

- ldap.jenkins.io for the controller
- pkg.origin.jenkins.io and repo.jenkins-ci.org for agents

PR #383, opened 2023-06-02
fix(trusted) use existing DNS zone for bounce record

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3486

There already was a public DNS zone acting as authority for `*.trusted.ci.jenkins.io` so better using it to make the `bounce` IN `A` record to be visible for public DNS clients


dduportal (commented):

PR #385, opened 2023-06-05
feat: add LDAP resource group and storage account

lemeurherve:
This PR creates a new storage account in its own resource group to store LDAP backups (LDIF dumps) instead of the existing `prodldap` one in the `prodldap` resource group, with network access limited to the admin IPs and to the publick8s cluster.

Ref: https://github.com/jenkins-infra/helpdesk/issues/3351#issuecomment-1576741071


dduportal (approved):
LGTM 👍

dduportal (approved):

dduportal:
This PR failed to be applied for multiple reasons: some where and are catcheable through the usual `terraform plan` of the checks, some are a bit more complicated.

- Name of storage accounts is constrained:
  - Only lower case letters and number
  - Between 3 and 24 characters
  - Must be unique across ALL Azure (not only our subscription, need to run `terraform apply` so view the error 🤦 )

- The creation of the storage account (once properly named, see above) fails with a weird HTTP/404 not found.
  - Found https://github.com/hashicorp/terraform-provider-azurerm/issues/12470 describing the issue
  - The use of an inline `network_rule` block (instead of an `azurerm_storage_account_network_rules` association) hides the root problem: the storage account is created, but the HTTP/404 appears when the Azure provider tries to update it to set up the network rules.
  - Root cause is because there are no storage endpoint for the subnet. Good thing we catch this now 😅
    ```text
      Error: updating Azure Storage Account network_rules "ldapjenkinsiobackups": storage.AccountsClient#Update: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="NetworkAclsValidationFailure" Message="Validation of network acls failure: SubnetsHaveNoServiceEndpointsConfigured:Subnets publick8s-tier of virtual network /subscriptions/<redacted>/resourceGroups/public/providers/Microsoft.Network/virtualNetworks/public-vnet do not have ServiceEndpoints for Microsoft.Storage resources configured. Add Microsoft.Storage to subnet's ServiceEndpoints collection before trying to ACL Microsoft.Storage resources to these subnets.."
    ```

PR #400, opened 2023-06-07
feat(ldap): add public IPv4

lemeurherve:
This PR adds a public IPv4 dedicated to LDAP which can be used to expose the service from `publick8s`, cf suggestion from @dduportal in https://github.com/jenkins-infra/kubernetes-management/pull/4044#discussion_r1220996916

PR #402, opened 2023-06-07
fix(ldap/public8s): move LDAP dedicated IPv4 definition from `ldap.jenkins.io.tf` to `publick8s.tf`

lemeurherve:
As https://github.com/jenkins-infra/kubernetes-management/pull/4044 failed as the IP created in #400 should have been in the `publick8s` cluster's node resource group instead of the `ldap` one, this PR moves its definition and output from LDAP resources definition to `publick8s` resources definition (coupled)

Fixup of #400


lemeurherve:
Terraform output:
> ldap_jenkins_io_ipv4_address = "20.10.205.3"

PR #406, opened 2023-06-13
fix(ldap,trusted) use the new public_ip resource for the LDAP connection from trusted.ci

dduportal:
As per https://github.com/jenkins-infra/helpdesk/issues/3622, the controller trusted.ci.jenkins.io is having issues.

These issues are caused to connections timing out when trying to reach ldap.jenkins.io in the 636 port:

```
Jun 13 06:52:27 controller docker-jenkins[8264]: Caused: org.springframework.ldap.CommunicationException: ldap.jenkins.io:636; nested exception is javax.naming.CommunicationException: ldap.jenkins.io:636 [Root exception is java.net.SocketTimeoutException: connect timed out]
```

It's because we migrated the LDAP which changed its IP.

Thanks to https://github.com/jenkins-infra/azure/pull/400, the LDAP external IP is now part of the Terraform resources: let's use it to ensure that NSGs of trusted.ci.jenkins.io are always up to date.

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -46,7 +46,6 @@ locals {
 
   external_services = {
     "puppet.jenkins.io"  = azurerm_public_ip.puppet_jenkins_io.ip_address
-    "ldap.jenkins.io"    = "52.184.219.77"
     "updates.jenkins.io" = "52.202.51.185"
   }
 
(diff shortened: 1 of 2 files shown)


dduportal (commented):
`Plan: 0 to add, 1 to change, 0 to destroy.` The plan is exactly what I expected

PR #418, opened 2023-06-29
feat(publick8s): add a new arm64 pool with taint

smerleCB:
as per https://github.com/jenkins-infra/helpdesk/issues/3623

imply to add toleration to use this new arm64 node pool as per : https://github.com/jenkins-infra/kubernetes-management/pull/4120

### DIFF
--- a/publick8s.tf
+++ b/publick8s.tf
@@ -109,6 +109,23 @@ resource "azurerm_kubernetes_cluster_node_pool" "arm64small" {
   tags                  = local.default_tags
 }
 
+resource "azurerm_kubernetes_cluster_node_pool" "arm64small_20230629" {
+  name                  = "arm64small-20230629"
+  vm_size               = "Standard_D4pds_v5" # 4 vCPU, 16 GB RAM, local disk: 150 GB and 19000 IOPS
+  os_disk_type          = "Ephemeral"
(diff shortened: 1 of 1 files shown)


publick8s.tf
```diff
@@ -109,6 +109,23 @@ resource "azurerm_kubernetes_cluster_node_pool" "arm64small" {
...
+  os_disk_size_gb       = 150 # Ref. Cache storage size at https://learn.microsoft.com/en-us/azure/virtual-machines/dpsv5-dpdsv5-series#dpdsv5-series (depends on the instance size)
+  kubernetes_cluster_id = azurerm_kubernetes_cluster.publick8s.id
+  enable_auto_scaling   = true
+  min_count             = 0
+  max_count             = 10
+  zones                 = [1]
+  vnet_subnet_id        = data.azurerm_subnet.publick8s_tier.id
+  tags                  = local.default_tags
+  node_taints = [
+    "pool=small:NoSchedule",
```
dduportal, line 125:
```suggestion
    "pool=small:NoSchedule",
    "arch=arm64:NoSchedule",
```

The "small" taint is interesting but we need one with the arch to forbid any Intel pod to be scheduled accidentally on this nodepool

dduportal (changes requested):

PR #421, opened 2023-07-04
breaking(ci.jenkins.io) bootstrap the new Gen2 VM

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3535

- Adds 2 DNS A records to reach the VM (without changing ci.jenkins.io, yet)
- Rename resources to stick to the "controller" naming like we did with trusted.ci (the proposal to make a module from @timja makes sense for a controller setup as we instantiate it 3 times, so let's get trusted and CI close in order to make the module happen later)
- Set up a first set of NSG rules to controler outbound


dduportal (commented):
`Plan: 13 to add, 2 to change, 0 to destroy.` ready to roll!

PR #431, opened 2023-07-07
fix(publick8s) specify pod subnets explicitly

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3582#issuecomment-1625011472


lemeurherve:
Without this, pods were allocated an IP in the "public" vnet but not specifically in the "publick8s-tier" subnet. (Some were in 10.244.x.x like the accountapp, which prevented it to access the ldap due to ldap loadBalancerSourceRanges not including it)

lemeurherve (approved):

dduportal (commented):
`Plan: 4 to add, 1 to change, 4 to destroy.` /o\

PR #433, opened 2023-07-10
feat(locks) add locks on all public IPs that should not change

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3582#issuecomment-1629210833

Notes:

- The public IPs on trusted.ci (for the inbound SSH to bounce VM and for the subnet's gateway) are not required to be locked
- Additional locks might be added to the data disks of ci.j or trusted.ci.j for instance, as part of https://github.com/jenkins-infra/helpdesk/issues/3479 if this works as expected


lemeurherve (approved):
🚢

PR #434, opened 2023-07-11
cleanup(ci.jenkins.io) remove leftovers from the former Gen1 VM

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3535#issuecomment-1629312380

This PR removes:

- The storage account `cijenkinsiovmagents` which was used by the Azure VM plugin of the former VM
- The resource group `eastus-cijenkinsio` which was used by the Azure VM and ACI plugins of the former VM


timja (approved):

lemeurherve (approved):

dduportal (commented):
```
Plan: 0 to add, 0 to change, 5 to destroy.
```

as expected

PR #436, opened 2023-07-12
feat(ci.jenkins.io) allow agents egress with HKP protocol

dduportal:
This PR allows ci.jenkins.io agents to reach OpenPGP keyserver using the [HKP](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=hkp) protocol.

The destination servers are restricted to the know usages:

- Ideally we should embed GPG keys in the repositories such as https://github.com/jenkins-infra/packer-images/tree/main/gpg-keys
- We want an ehxaustive list of destination to limit the tentatives of retrieving bad keys

It's used by some containers in the jenkinsci/acceptance-test-harness (ATH) builds.

Related to https://github.com/jenkins-infra/helpdesk/issues/3664


dduportal:
> (it probably should have been communicated that ports were being restricted in the first place unless I missed something)

Good point. Let me send an email!


> seems pointless to restrict the destination for CI, keyserver.ubuntu adds a new IP address and then suddenly ATH breaks until we remember about this?

The idea is to have an automated process updating the IPs at least once a day to avoid this breakage. Same as the webhooks outbound IPs of GitHub.


I would advocate that the ATH breaking for this is a good thing:

- It let us have more knowledge of what is running in our polatform and define it with automated and auditable way (otherwise if the SRE infra team has to dig down on each repo of jenkinsci/, that won't be sustainable)
- We can start the discussion about fixing the ATH properly by embedding the GPG key in the repo and not only avoid opening the HKP outbound protocol, but also avoiding any Mitm in the GPG key retrieval (not mentioning the depreciation of keyservers)

dduportal:
(PR checks are failing due another error fixed in #437)

lemeurherve (approved):

lemeurherve:
> keyserver.ubuntu adds a new IP address and then suddenly ATH breaks until we remember about this?

@timja I've added an updatecli manifest to keep the list of its IPs up to date.

lemeurherve (approved):

PR #445, opened 2023-08-07
feat: add `updates.jenkins.io` storage account, redis database and CNAME record

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/2649

### DIFF
--- a/updates.jenkins.io.tf
+++ b/updates.jenkins.io.tf
@@ -42,7 +42,7 @@ output "updates_jenkins_io_database_hostname" {
 
 output "updates_jenkins_io_database_primary_access_key" {
   sensitive = true
-  value     = azurerm_redis_cache.updates_jenkins_io_files.primary_access_key
+  value     = azurerm_redis_cache.updates_jenkins_io_database.primary_access_key
 }
 
 # Service DNS record
--- a/updates.jenkins.io.tf
+++ b/updates.jenkins.io.tf
@@ -5,7 +5,7 @@ resource "azurerm_resource_group" "updates_jenkins_io" {
   tags     = local.default_tags
 }
 
-resource "azurerm_storage_account" "updates_jenkins_io_files" {
+resource "azurerm_storage_account" "updates_jenkins_io" {
   name                          = "updatesjenkinsiofiles"
   resource_group_name           = azurerm_resource_group.updates_jenkins_io.name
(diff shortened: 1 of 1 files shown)


lemeurherve:
> Plan: 4 to add, 0 to change, 0 to destroy.

updates.jenkins.io.tf
```diff
@@ -0,0 +1,56 @@
+# Storage account
+resource "azurerm_resource_group" "updates_jenkins_io" {
+  name     = "updates-jenkins-io"
+  location = var.location
+  tags     = local.default_tags
+}
+
+resource "azurerm_storage_account" "updates_jenkins_io_files" {
+  name                          = "updatesjenkinsiofiles"
```
dduportal, line 9:
```suggestion
  name                          = "updatesjenkinsio"
```

=> A storage account can have different services: files, blob storages, etc.

updates.jenkins.io.tf
```diff
@@ -0,0 +1,56 @@
...
+  resource_group_name           = azurerm_resource_group.updates_jenkins_io.name
+  location                      = azurerm_resource_group.updates_jenkins_io.location
+  account_tier                  = "Standard"
+  account_replication_type      = "GRS"    # recommended for backups
```
dduportal, line 13:
```suggestion
  account_replication_type      = "LRS"    # recommended for backups
```

Note sure why you would want to use GRS? https://learn.microsoft.com/en-us/azure/storage/common/storage-redundancy states that LRS are more cost effective and it's not clear why we would want to georeplicate?

updates.jenkins.io.tf
```diff
@@ -0,0 +1,56 @@
...
+  tags = local.default_tags
+}
+
+output "updates_jenkins_io_files_primary_access_key" {
+  sensitive = true
+  value     = azurerm_storage_account.updates_jenkins_io_files.primary_access_key
+}
+
+# Redis database
+resource "azurerm_redis_cache" "updates_jenkins_io_database" {
```
dduportal, line 26:
```suggestion
resource "azurerm_redis_cache" "updates_jenkins_io" {
```

The name "database" is already implied by the fact it is a Redis resource

dduportal (changes requested):
First pass of review.

- Naming is not blocking
- Using `GRS` sounds like a problem at first sight though

PR #450, opened 2023-08-17
feat(ci.jenkins.io, trusted.ci.jenkins.io) use a module for the controller resources

dduportal:
This is a preliminary work for https://github.com/jenkins-infra/helpdesk/issues/3688 (to allows cert.ci VM to be reconstructed properly on a gen2 VM).

This PR introduces the use of a terraform module to define a "Jenkins Azure Controller" so our 3 controllers would be easier to manage with the same naming conventions and rules.

A few notes:

- This is a first version of the module with only a small subset of resources (no NSG, no ephemeral agent resources, no Azure AD resources... yet).
- The 'moved' directive have been used to avoid destroying resources in complement with a backward compatibility around naming has been implemented by a renaming of resources for trusted.ci is required

### DIFF
--- a/trusted.ci.jenkins.io.tf
+++ b/trusted.ci.jenkins.io.tf
@@ -21,32 +21,59 @@ data "azurerm_subnet" "trusted_ephemeral_agents" {
   resource_group_name  = data.azurerm_resource_group.trusted.name
   virtual_network_name = data.azurerm_virtual_network.trusted.name
 }
-resource "azurerm_resource_group" "trusted_ci_jenkins_io_ephemeral_agents" {
-  name     = "jenkinsinfra-trusted-ephemeral-agents"
-  location = var.location
-  tags     = local.default_tags
+####################################################################################
+## Resources for the Controller VM
+####################################################################################
+module "trusted_ci_jenkins_io" {
+  source = "./.shared-tools/terraform/modules/azure-jenkins-controller"
+
+  controller_short_hostname    = "trusted.ci"
+  location                     = data.azurerm_virtual_network.trusted.location
+  admin_username               = local.admin_username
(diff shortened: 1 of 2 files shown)


dduportal:

- Thanks to the `moved` directives, than will be removable once applied, this change does not destroy any resource
- The new resource will be a DNS record for `controller.trusted.ci.jenkins.io` in the private DNS zone

trusted.ci.jenkins.io.tf
```diff
@@ -21,32 +21,59 @@ data "azurerm_subnet" "trusted_ephemeral_agents" {
...
+  dns_zone_name                = azurerm_private_dns_zone.trusted.name
+  dns_resourcegroup_name       = data.azurerm_resource_group.trusted.name
+  controller_subnet_id         = data.azurerm_subnet.trusted_ci_controller.id
+  controller_data_disk_size_gb = 128
+  controller_vm_size           = "Standard_D2as_v5"
+  default_tags                 = local.default_tags
+
+  controller_resourcegroup_name = "jenkinsinfra-trusted-ci-controller"
+  controller_vm_name            = ""
```
timja, line 42: why is this set to an empty string?

trusted.ci.jenkins.io.tf
```diff
@@ -21,32 +21,59 @@ data "azurerm_subnet" "trusted_ephemeral_agents" {
...
-  tags     = local.default_tags
+####################################################################################
+## Resources for the Controller VM
+####################################################################################
+module "trusted_ci_jenkins_io" {
+  source = "./.shared-tools/terraform/modules/azure-jenkins-controller"
+
+  controller_short_hostname    = "trusted.ci"
+  location                     = data.azurerm_virtual_network.trusted.location
+  admin_username               = local.admin_username
```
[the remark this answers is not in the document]
dduportal: This value is also defined for other VMs different from controllers, hence passing it. Adding a default create a duplication of code with the `locals.tf` value in this project, hence the choice to pass it as a parameter

trusted.ci.jenkins.io.tf
dduportal: https://github.com/jenkins-infra/azure/pull/452

PR #452, opened 2023-08-17
fix(ci.jenkins.io, trusted.ci.jenkins.io) fix controler module initial usage

dduportal:
Follow up https://github.com/jenkins-infra/azure/pull/450#discussion_r1296854845

It also fixes the error detected during the apply of #450:

```
Error: creating/updating DNS A Record "controller.trusted.ci" (Zone "trusted.ci.jenkins.io" / Resource Group "trusted"): unexpected status 404 with error: ParentResourceNotFound: Failed to perform 'write' on resource(s) of type 'dnszones/A', because the parent resource '/subscriptions/****/resourceGroups/trusted/providers/Microsoft.Network/dnszones/trusted.ci.jenkins.io' could not be found.
```

### DIFF
--- a/ci.jenkins.io.tf
+++ b/ci.jenkins.io.tf
@@ -50,6 +50,10 @@ moved {
   to   = module.ci_jenkins_io.azurerm_dns_a_record.controller
 }
 moved {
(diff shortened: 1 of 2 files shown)


dduportal (commented):

PR #456, opened 2023-08-18
feat(ci.jenkins.io, trusted.ci.jenkins.io) manage ephemeral agents (azure-vm plugin) resources with the Terraform module

dduportal:
This change was already applied manually (to ensure changes are ok due to NSG rules)

PR #460, opened 2023-08-24
cleanup(trusted.ci, ci.jenkins.io) move privatevpn NSGs and remove unused directives

dduportal:
Follow up of #456 this PR does 2 things:

- Removal of unused resources (duplicated/commented/etc.)
- Moving the NSG rule allowing SSH from private VPN network to controller and ephemeral agents to the TF module


dduportal:
Plan says: `Plan: 4 to add, 0 to change, 5 to destroy.` which is expected:

- Removal of 1 unused NSG rule `allow_inbound_jenkins_usage_from_vnet_to_controller`
- 2 NSG rule per TF module instanciations are renamed, hence destroyed and re-created, eg.g. 4 add and 4 destroys

PR #461, opened 2023-08-24
feat(cert.ci.jenkins.io) create a new controller VM and associated resources in the new network

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3688

A few notes:

- The VM size changed by following the Azure advisory recommandation (except it is a `_v5` instead of `_v4` which is the same as trusted.ci.jenkins.io):

Capture d’écran 2023-08-17 à 10 21 25

- Data disk set to 128 like trusted.ci.jenkins.io as the current VM does not use a lot of data in the Jenkins home:

```
$ df -h /var/lib/jenkins/
Filesystem                Size  Used Avail Use% Mounted on
/dev/mapper/jenkins-home  295G   27G  253G  10% /var/lib/jenkins
```

- The new terraform module is used to have the same kind of resources (NSG, VM, etc.) as for ci.jenkins.io and trusted.ci.jenkins.io

### DIFF
--- a/cert.ci.jenkins.io.tf
+++ b/cert.ci.jenkins.io.tf
@@ -1,3 +1,50 @@
+# Data of resources defined in https://github.com/jenkins-infra/azure-net
+data "azurerm_resource_group" "cert_ci_jenkins_io" {
+  name = "cert-ci-jenkins-io"
+}
+
+module "cert_ci_jenkins_io" {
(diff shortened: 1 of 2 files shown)


dduportal (commented):
`Plan: 39 to add, 0 to change, 0 to destroy.` and no mentions of `trusted` LGTM

PR #479, opened 2023-09-23
chore(updatecli) track terraform providers with native updatecli resource

dduportal:
Since `updatecli` [v0.60.0](https://github.com/updatecli/updatecli/releases/tag/v0.60.0), there are a new native resource available to track terraform providers from the hashicorp registry to the terraform lock files.

This PR fixes the broken updatecli builds of this repository and switches to native terraform resources which includes:

- One manifest per provider is added
- Removed the "all in one" updater and the former script
- No need to use a custom hashicorp agent image to execute updatecli as it's native in updatecli now (no docker, no terraform required)


timja (approved):
renovate can do this natively without having to provide all this config =/

dduportal:

dduportal:
> renovate can do this natively without having to provide all this config =/

Yup, we know but we still want to keep using updatecli.

PR #487, opened 2023-09-27
chore(updates.jenkins.io): add NS record for `cloudflare.jenkins.io`

lemeurherve:
Trying something to help resolving edge certificate issue on CloudFlare for westeurope.cloudflare.jenkins.io zone.


dduportal (approved):
Approving to unblock, but as we discussed let's wait for Cloudlfare to take a look

dduportal:
Closing as per Cloudlfare's support response: our problem is related to the `CAA` records, so no need to create records which won't be used.

PR #490, opened 2023-09-29
chore: use shared admins from shared-tools module

dduportal:
Twin of of https://github.com/jenkins-infra/digitalocean/pull/165

This PR is a first try at sharing a set common values across our terraform projects.

A new [data-only Terraform module](https://developer.hashicorp.com/terraform/language/modules/develop/composition#data-only-modules) has been added in the jenkins-infra/shared-tools to hold the shared values: https://github.com/jenkins-infra/shared-tools/commit/36589078701232007dbdf8b5034fd93901d072c3.

The following changes are introduced:

- the admin IPs from the shared module are used instead of the locals ones. Ping @smerle33 @lemeurherve can you check the shared module and terraform plan maps to your public IPs (or update them on https://github.com/jenkins-infra/shared-tools/blob/main/terraform/modules/jenkins-infra-shared-data/outputs.tf)

💡 Notes:

- This module, for now, only specify a set of outputs providing the shared data to projects. It may provide data sources (Azure DNS zones or network ids?) in the future if need be for instance.
- The choice of mixing IPv6 and IPv4 in the string arrays is because most of the usages requires both. If we need to filter by type, we can use `can(cidrnetmask(...)` and `!can(cidrnetmask(...))` (ref. https://developer.hashicorp.com/terraform/language/functions/cidrnetmask)
- The idea is that, in the near future, each terraform project should be able to update the values of the module when needed (case of public or outbound IPs of our services) or a human (admin IPs).

### DIFF
--- a/archives.tf
+++ b/archives.tf
@@ -15,8 +15,10 @@ resource "azurerm_storage_account" "archives" {
   min_tls_version                   = "TLS1_2" # default value, needed for tfsec
 
   network_rules {
-    default_action             = "Deny"
-    ip_rules                   = values(local.admin_allowed_ips)
+    default_action = "Deny"
+    ip_rules = flatten(concat(
+      [for key, value in module.jenkins_infra.admin_public_ips : value]
+    ))
     virtual_network_subnet_ids = [data.azurerm_subnet.privatek8s_tier.id]
     bypass                     = ["AzureServices"]
   }
--- a/archives.tf
+++ b/archives.tf
@@ -17,7 +17,7 @@ resource "azurerm_storage_account" "archives" {
   network_rules {
     default_action = "Deny"
     ip_rules = flatten(concat(
-      [for key, value in module.jenkins_infra.admin_public_ips : value]
+      [for key, value in module.jenkins_infra_shared_data.admin_public_ips : value]
     ))
     virtual_network_subnet_ids = [data.azurerm_subnet.privatek8s_tier.id]
     bypass                     = ["AzureServices"]
--- a/ldap.jenkins.io.tf
+++ b/ldap.jenkins.io.tf
@@ -22,7 +22,7 @@ resource "azurerm_storage_account_network_rules" "ldap_access" {
 
   default_action = "Deny"
   ip_rules = flatten(concat(
-    [for key, value in module.jenkins_infra.admin_public_ips : value]
+    [for key, value in module.jenkins_infra_shared_data.admin_public_ips : value]
   ))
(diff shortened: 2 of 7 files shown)


main.tf
```diff
@@ -7,3 +7,7 @@ data "azuread_service_principal" "terraform_production" {
 # Data source used to retrieve the subscription id
 data "azurerm_subscription" "jenkins" {
 }
+
+module "jenkins_infra" {
```
lemeurherve, line 11:
```suggestion
module "jenkins_infra_shared_data" {
```

archives.tf
```diff
@@ -15,8 +15,10 @@ resource "azurerm_storage_account" "archives" {
   min_tls_version                   = "TLS1_2" # default value, needed for tfsec
 
   network_rules {
-    default_action             = "Deny"
-    ip_rules                   = values(local.admin_allowed_ips)
+    default_action = "Deny"
+    ip_rules = flatten(concat(
+      [for key, value in module.jenkins_infra.admin_public_ips : value]
```
lemeurherve, line 20:
```suggestion
      [for key, value in module.jenkins_infra_shared_data.admin_public_ips : value]
```

ldap.jenkins.io.tf
```diff
@@ -20,8 +20,10 @@ resource "azurerm_storage_account" "ldap_backups" {
 resource "azurerm_storage_account_network_rules" "ldap_access" {
   storage_account_id = azurerm_storage_account.ldap_backups.id
 
-  default_action             = "Deny"
-  ip_rules                   = values(local.admin_allowed_ips)
+  default_action = "Deny"
+  ip_rules = flatten(concat(
+    [for key, value in module.jenkins_infra.admin_public_ips : value]
```
lemeurherve, line 25:
```suggestion
    [for key, value in module.jenkins_infra_shared_data.admin_public_ips : value]
```

privatek8s.tf
```diff
@@ -31,7 +31,14 @@ resource "azurerm_kubernetes_cluster" "privatek8s" {
 
   api_server_access_profile {
     authorized_ip_ranges = setunion(
-      formatlist("%s/32", values(local.admin_allowed_ips)),
+      formatlist(
+        "%s/32",
+        flatten(
+          concat(
+            [for key, value in module.jenkins_infra.admin_public_ips : value]
```
lemeurherve, line 38:
```suggestion
            [for key, value in module.jenkins_infra_shared_data.admin_public_ips : value]
```

lemeurherve (approved):

lemeurherve (approved):

PR #492, opened 2023-10-06
feat(updates.jenkins.io): `updates-jenkins-io` Azure File Share SAS

lemeurherve:
This PR allows to retrieve the URL including the account SAS token needed to interact with the `updates-jenkins-io` file share with `azcopy` (we can't use Azure AD with File Share, nor we want on trusted).

Docs:
- https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10?toc=%2Fazure%2Fstorage%2Fblobs%2Ftoc.json&bc=%2Fazure%2Fstorage%2Fblobs%2Fbreadcrumb%2Ftoc.json#authorize-azcopy
- https://learn.microsoft.com/en-us/rest/api/storageservices/create-account-sas

Ref: https://github.com/jenkins-infra/helpdesk/issues/2649

### DIFF
--- a/updates.jenkins.io.tf
+++ b/updates.jenkins.io.tf
@@ -23,6 +23,48 @@ resource "azurerm_storage_share" "updates_jenkins_io" {
   quota                = 2 # updates.jenkins.io total size in /www/updates.jenkins.io: 400Mo (Mid 2023)
 }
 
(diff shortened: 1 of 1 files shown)


dduportal (approved):
```
Changes to Outputs:
  + updates_jenkins_io_sas_url_query_string = (sensitive value)
```

LGTM

PR #493, opened 2023-10-06
chore(updates.jenkins.io): include Azure File Storage connection string in output

lemeurherve:
Follow-up of #492


lemeurherve:
It seems that terraform doesn't like concatenate sensitive values. This PR wasn't that important anyway.

PR #494, opened 2023-10-09
fix(updates.jenkins.io): add service as resource type for the SAS token

lemeurherve:
This PR adds `service` as resource type so the account SAS token is accepted for commands like `azcopy list <file-share-url-with-sas-token>`.

It also adds a sensitive output to retrieve the File Share connection string (QoL).

Tested by manually generating tokens from the Azure Portal. (The added `signed_version` comes from these tests)


lemeurherve:
Changes to outputs only:
```

Changes to Outputs:
  + updates_jenkins_io_primary_connection_string = (sensitive value)
  # Warning: this attribute value will be marked as sensitive and will not
  # display in UI output after applying this change.
  ~ updates_jenkins_io_sas_url_query_string      = (sensitive value)

You can apply this plan to save these new output values to the Terraform
state, without changing any real infrastructure.
```

PR #495, opened 2023-10-09
chore(updates.jenkins.io): more useful File Share outputs

lemeurherve:
This PR adds an output for the File Share URL, which is needed in combination with the SAS token query string to make requests via `azcopy`.

It also removes the `updates_jenkins_io_storage_account_primary_access_key` output, not needed.

Follow-up of #493 & #494

### DIFF
--- a/updates.jenkins.io.tf
+++ b/updates.jenkins.io.tf
@@ -61,19 +61,14 @@ data "azurerm_storage_account_sas" "updates_jenkins_io" {
   }
 }
 
-output "updates_jenkins_io_primary_connection_string" {
+output "updates_jenkins_io_share_url" {
   sensitive = true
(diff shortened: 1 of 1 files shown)


lemeurherve:
```
Changes to Outputs:
  - updates_jenkins_io_primary_connection_string          = (sensitive value)
  + updates_jenkins_io_sas_query_string                   = (sensitive value)
  - updates_jenkins_io_sas_url_query_string               = (sensitive value)
  + updates_jenkins_io_share_url                          = (sensitive value)
  - updates_jenkins_io_storage_account_primary_access_key = (sensitive value)

You can apply this plan to save these new output values to the Terraform
state, without changing any real infrastructure.
```

Self-merging, concerns only outputs.

PR #497, opened 2023-10-10
feat(matomo) add a MySQL database on the public-db instance

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3602

This PR adds a managed MySQL database for matomo with an associated user and password.

The grants are also applied to this user as per https://matomo.org/faq/how-to-install/faq_23484/. Note that the `FILE` grant is not added because it would be global to the `public-db` instance while we're not even sure it is needed (or if the mentionned file load extension is present on Azure flexible instances)


smerle33 (approved):
🚀

PR #508, opened 2023-11-07
fix(publick8s) ensure trusted authorized IP uses a CIDR notation

dduportal:
This PR avoids to update the cluster on each `terraform apply`


dduportal (commented):
`No changes. Your infrastructure matches the configuration.` Yes !

Self merging

PR #517, opened 2023-11-21
feat(publick8s) add a new arm64 node pool in the compute zone 3 

dduportal:
This PR adds a new node pool in the same compute zone (`3`) as the other node pools used for Intel.

It helps to fix https://github.com/jenkins-infra/kubernetes-management/pull/4698 which shows there is an affinity error between jenkins-weekly's PVC (`useast2-3`) and the current `arm64small2` node pool machines (`useast2-1`) leading to nonscheduled pod.

Note: we kept the same machine sizes as it defaults to 3 machines which is cheaper than 2 "medium sized" machines (medium VM is 2x the time of a small one)

### DIFF
--- a/publick8s.tf
+++ b/publick8s.tf
@@ -23,6 +23,10 @@ data "azurerm_subnet" "public_vnet_data_tier" {
   virtual_network_name = data.azurerm_virtual_network.public.name
 }
 
+locals {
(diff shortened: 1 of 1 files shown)


smerle33 (approved):
🚀

PR #518, opened 2023-11-21
feat(publick8s) remove unused or invalid node pools

dduportal:


dduportal (commented):
Self merging (as part of the pair session with @smerle33 )

PR #528, opened 2023-11-27
chore(contributors.jenkins.io): recreate Storage Account and File Share

lemeurherve:
This PR recreates the storage account and the file share for contributors.jenkins.io

The network rules deleted in a hotfix commit were still present when looking at the storage account in Azure Portal.
I've deleted the (empty) storage account to recreate it with this PR.

Other changes:
- ZRS: no need for GRS, ZRS is sufficient
- Type StorageV2: default value from the terraform provider, "Storage" (v1) is the legacy one, no cost change between them.
- Add back network rules, allowing our IPs and the ones from publick8s and privatek8s vnets.

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/3809#issuecomment-1822357561

### DIFF
--- a/contributors.jenkins.io.tf
+++ b/contributors.jenkins.io.tf
@@ -13,10 +13,19 @@ resource "azurerm_storage_account" "contributors_jenkins_io" {
   resource_group_name       = azurerm_resource_group.contributors_jenkins_io.name
   location                  = azurerm_resource_group.contributors_jenkins_io.location
   account_tier              = "Standard"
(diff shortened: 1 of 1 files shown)


smerle33 (approved):
LGTM with pairing

lemeurherve:
We disabled the terraform job, deleted the storage account via the Azure Portal, ran `terraform plan` locally to ensure everything was correct, merged the PR then reactivated the terraform job.

PR #529, opened 2023-11-28
chore(contributors.jenkins.io): add SAS token output

lemeurherve:
This PR adds outputs to retrieve fine grained File Share SAS token so we can use azcopy instead of blobxfer and entire Storage Account access key.

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/3809#issuecomment-1822357561


dduportal (approved):

lemeurherve:
> Changes to Outputs:
>   + contributors_jenkins_io_sas_query_string = (sensitive value)
>   + contributors_jenkins_io_share_url        = (sensitive value)
> 
> You can apply this plan to save these new output values to the Terraform
> state, without changing any real infrastructure.

PR #532, opened 2023-11-29
cleanup: remove unused ACI resources for trusted.ci and cert.ci

dduportal:
While working on the Terraform module split in https://github.com/jenkins-infra/helpdesk/issues/3818, it appeared that some unneeded (and unused) resources were creating by the "former module".

The migration to the "3 distinct modules" kept the resources, but this PR removes them as we won't need to use ACI with trusted and cert.


lemeurherve (approved):

dduportal:
`Plan: 0 to add, 0 to change, 4 to destroy.`

PR #539, opened 2023-12-05
Revert "feat(publick8s): add a new systempool with taint"

smerle33:
Reverts jenkins-infra/azure#537

as an error occurred : 
```
Agent Pool Name: "systempool3"): agentpools.AgentPoolsClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: Code="SystemPoolHasRestrictedTaint" Message="Placing custom taints on system pool is not supported(except 'CriticalAddonsOnly' taint or taint effect is 'PreferNoSchedule'). Please refer to https://aka.ms/aks/system-taints for detail"
```

PR #540, opened 2023-12-05
fixup(publick8s): add the correct taint to the new system pool

smerle33:
correct taint as per https://learn.microsoft.com/en-us/azure/aks/use-system-pools?tabs=azure-cli#system-and-user-node-pools

follown up #539


smerle33:
Plan: 1 to add, 0 to change, 0 to destroy.

PR #546, opened 2023-12-12
hotfix(ci.jenkins.io) ensure ACI agents are reusing the same role (as…

dduportal:
… its scoped across all subscriptions)


dduportal:
Self merging as hotfix to unblock main branch failures. Gotta update the body after once validated.

PR #548, opened 2023-12-14
feat(trusted.ci.jenkins.io) add resources to allow spawning azure VM agents in the secondary subscription

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3818

This PR adds the required resources to spawn azure VM agents in the secondary subscription for trusted.ci.jenkins.io.

### DIFF
--- a/trusted.ci.jenkins.io.tf
+++ b/trusted.ci.jenkins.io.tf
@@ -2,9 +2,18 @@
 data "azurerm_resource_group" "trusted_ci_jenkins_io" {
   name = "trusted-ci-jenkins-io"
 }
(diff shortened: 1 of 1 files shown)


dduportal:
`Plan: 19 to add, 0 to change, 0 to destroy.` => as expected

timja (approved):
the custom role seems overkill, Reader should be just fine there. rest looks fine.

dduportal:
> the custom role seems overkill, Reader should be just fine there. rest looks fine.

The intent is to not give controllers permissions they don't need. The `Reader` built-in role give access to uneeded resources when applied to a wide scope.
With this custom role, it allows controller to have full "Contributor" in its resource group but nowhere else

PR #554, opened 2023-12-18
feat(infra.ci.jenkins.io) add secondary Azure VM subscription resources

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3818#issuecomment-1860472394

Expecting 19 new resources to be created (same as previous PRs for other controllers)

PR #555, opened 2023-12-18
fix(infra.ci.jenkins.io) fixup of #554 by using correct SP ID and scopes for role assignements

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3818#issuecomment-1860472394

Fixup of #554 , this PRs fixes the failing build on the main branch (not caught by `terraform plan` because it is an HTTP/400)


smerle33 (approved):
🚀

dduportal (commented):
`Plan: 4 to add, 0 to change, 1 to destroy.`

- 1 `role` to re-create
- 3 role assignment which failed to be created with the previous PR

PR #556, opened 2023-12-19
feat(packer): move packer-builds to the new provider azure sponsored

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3818

this PR moves the packer-builds resources groups to the new subscription.

the goal is too fix build errors on jenkins-infra/packer-images for windows machines (winRM timeout)

### DIFF
--- a/packer-resources.tf
+++ b/packer-resources.tf
@@ -46,6 +46,7 @@ resource "azurerm_resource_group" "packer_images" {
 }
 
 resource "azurerm_resource_group" "packer_builds" {
+  provider = azurerm.jenkins-sponsorship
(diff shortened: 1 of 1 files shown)


dduportal:
Putting as draft while we pair on locating the packer network settings (to ensure new packer VMs are built in the new subscriptions)

dduportal (approved):
- `Plan: 6 to add, 0 to change, 3 to destroy.`
- Fun: it looks like that adding the provider does not mark them to be tainted. Instead Terraform tells us it will create the new one (in the new provider) from scratch and the former RGs will be orphans. Good to know 🤔

PR #562, opened 2023-12-21
chore: move datasource definitions in their associated TF files

dduportal:
This PR should not make any resource change: it's only a "cleanup".

It moves the network data sources (vnet RGs, vnets and subnets)  to the `vnet.tf` and DNS data sources to `dns.tf` to ensure we have the same convention az in azure-net despite physical separation of projects


dduportal (commented):
`No changes. Your infrastructure matches the configuration.` => self merging

PR #570, opened 2024-01-06
feat(updates.jenkins.io) restrict SA to only a few administrative IPs

dduportal:
This PR is related to both https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-1874074475 and https://github.com/jenkins-infra/helpdesk/issues/3875

It restricts the Azure SA used for updates.jenkins.io to only a few administrative IPs including trusted.ci's 2 networks.

It also allows trusted.ci new network (sponsored) agents to access publick8s controler (for update center)

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -31,8 +31,6 @@ locals {
     "keyserver.ubuntu.com" = ["162.213.33.8", "162.213.33.9"]
   }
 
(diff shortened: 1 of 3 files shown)


dduportal (commented):
`Plan: 0 to add, 2 to change, 0 to destroy.` with plan as expected: self merging

PR #571, opened 2024-01-06
hotfix(updates.jenkins.io) SA network restriction must have plain IPs and should not disable public access to apply restriction

dduportal:
Fixup of #570 

Related to https://github.com/jenkins-infra/helpdesk/issues/2649 and https://github.com/jenkins-infra/helpdesk/issues/3875


dduportal:
Self mergin (applied locally as a hot fix)

PR #579, opened 2024-01-17
feat(publick8s,privatek8s) set up outbound LB to support more SNAT connections

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3908

This PR tunes the network outbound method used for both `publick8s` and `privatek8s` (using loadbalancers) with:

- TCP idle timeout decrease from 30 min (default) to 4 min to recycle sockets way more often
- Force static allocation of `3200` (and `1600`)  port on the public outbound IPs as per the Azure Metrics (these values are the upper of each amount of SNAT connection diagrams). 
  - Note it disable the dynamic allocation: this should be a problem if we have more than 50 nodes per cluster. Not the case for these 2.

### DIFF
--- a/privatek8s.tf
+++ b/privatek8s.tf
@@ -44,8 +44,15 @@ resource "azurerm_kubernetes_cluster" "privatek8s" {
   }
 
   network_profile {
-    network_plugin = "azure"
-    network_policy = "azure"
+    network_plugin    = "azure"
+    network_policy    = "azure"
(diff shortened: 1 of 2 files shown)


dduportal (commented):

smerle33 (approved):
🚀

lemeurherve (approved):

PR #593, opened 2024-01-22
hotfix(ci.jenkins.io) migrate controller RG to the module

dduportal:
Fixup of #583 

This PR migrates the explicit RG definition for ci.jenkins.io controller in the sponsored subscription into the new module introduced in #583 

It also updates, along the way, references to new (properly named) module output `controller_service_principal_id` instead of `controler_service_principal_id`


dduportal:
Self merging to fix `main` branch status

`Plan: 18 to add, 0 to change, 0 to destroy.`

PR #595, opened 2024-01-23
chore(contributors.jenkins.io): add a stored access policy to the storage account

lemeurherve:
Cf https://learn.microsoft.com/en-us/azure/storage/common/storage-sas-overview#best-practices-when-using-sas, This PR defines a stored access policy which allows to easily revoke a SAS token by backdating its expiry date cf https://learn.microsoft.com/en-us/rest/api/storageservices/define-stored-access-policy:

> You can also use a stored access policy to revoke a signature after it has been issued.

We can then pass this stored access policy as argument to create a SAS token with `az`:
```
az storage share generate-sas --name updates-jenkins-io --account-name updatesjenkinsio  --https-only --permissions dlrw --policy-name contributorsjenkinsio-stored-access-policy --expiry 2024-01-24T00:00Z
```
(Haven't tested yet if the `permissions` parameter is needed)

Ref:

### DIFF
--- a/contributors.jenkins.io.tf
+++ b/contributors.jenkins.io.tf
@@ -35,8 +35,8 @@ resource "azurerm_storage_share" "contributors_jenkins_io" {
     id = "contributorsjenkinsio-stored-access-policy"
     access_policy {
       permissions = "rwdl"
-      start  = "2024-01-22T00:00:00Z"
-      expiry = "2024-04-21T00:00:00Z"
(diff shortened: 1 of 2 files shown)


lemeurherve:
> Plan: 0 to add, 1 to change, 0 to destroy.

lemeurherve:
Self-merging to continue on the subject, low risk as the last contributor spotlight in preparation has been published.

PR #601, opened 2024-01-24
fix(ci.jenkins.io) re-create the new VM (sponsored) with proper DNS

dduportal:
The goal of this PR is to recreate  the new (sponsored) VM and ensure it has a proper DNS (and hostname) so that Puppet can manage it properly.

Related to https://github.com/jenkins-infra/helpdesk/issues/3913

### DIFF
--- a/ci.jenkins.io.tf
+++ b/ci.jenkins.io.tf
@@ -52,8 +52,9 @@ module "ci_jenkins_io_sponsorship" {
     azuread     = azuread
   }
 
-  service_custom_name          = "ci.jenkins.io-sponsorship"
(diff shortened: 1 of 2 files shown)


dduportal (commented):
`Plan: 32 to add, 4 to change, 30 to destroy.`

- 2 changes are unrelated (node pool on privatek8s because changed manually + ACL file change)
- 2 changes are expected: Azure AD/Entra Application and a Azure Role Definition have their display names changed
- 30 resources destroyed as expected (the resources managed by the changed module except the 2 changed above)
- 32 creation: the 30 resources destroyed to b re-created + 2 DNS records which did not exist before

PR #602, opened 2024-01-24
fix(infra.ci.jenkins.io): correct scope on contributors.jenkins.io resource group

lemeurherve:
This PR sets the misnamed `file_share_id` (it should be renamed to `scope` in another PR) from contributors.jenkins.io File Share id to its resource group id.

It's used in the fileshare-writer module to define the scope of the role assignment given to the service principal.

Ref:

### DIFF
--- a/infra.ci.jenkins.io.tf
+++ b/infra.ci.jenkins.io.tf
@@ -76,7 +76,7 @@ module "infra_ci_jenkins_io_fileshare_serviceprincipal_writer" {
   active_directory_owners    = [data.azuread_service_principal.terraform_production.id]
   active_directory_url       = "https://github.com/jenkins-infra/azure"
   service_principal_end_date = "2024-06-20T23:00:00Z"
-  file_share_id              = azurerm_storage_share.contributors_jenkins_io.resource_manager_id
+  file_share_id              = azurerm_resource_group.contributors_jenkins_io.io
   default_tags               = local.default_tags
 }
 output "infra_ci_jenkins_io_fileshare_serviceprincipal_writer_id" {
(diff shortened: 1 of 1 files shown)


lemeurherve:
> Plan: 1 to add, 1 to change, 1 to destroy.

Recreating the role assignment as expected when changing the scope.

infra.ci.jenkins.io.tf
```diff
@@ -76,7 +76,7 @@ module "infra_ci_jenkins_io_fileshare_serviceprincipal_writer" {
   active_directory_owners    = [data.azuread_service_principal.terraform_production.id]
   active_directory_url       = "https://github.com/jenkins-infra/azure"
   service_principal_end_date = "2024-06-20T23:00:00Z"
-  file_share_id              = azurerm_storage_share.contributors_jenkins_io.resource_manager_id
+  file_share_id              = azurerm_resource_group.contributors_jenkins_io.id
```
dduportal, line 79: Scoping to resource group works and is ok for the time being but it looks like it is not restricting enough. 

WDYT about trying with a restriction at the SA level (in-between the filestorage and the resource group).

If you haven't tried, I don't mind checking later once you'll have set up this system (and I'll add a comment here subsequently)

lemeurherve:
It didn't worked anyway.

I'll try again with the File Share's `resource_manager_id` instead of its id like before this PR, as it's set for the updates.jenkins.io File Share: https://github.com/jenkins-infra/azure/blob/4aea9949038d69bd22bf5618a324ddb541f0b69e/trusted.ci.jenkins.io.tf#L84

lemeurherve:
Nevermind, it worked (previously too!):

image

I'm not sure which is better then, the File Share id or resource_manager_id... 🤔

lemeurherve:
https://github.com/jenkins-infra/azure/pull/603#issuecomment-1910126133

PR #609, opened 2024-01-26
feat(privatek8s): minimum count for autoscalling to 1 for spot instances nodes

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3918

for the ARM64 node the price for 1 node all month is around 14$:
﻿﻿image

for the intel node the price for 1 node all month is around 33$:
image

### DIFF
--- a/privatek8s.tf
+++ b/privatek8s.tf
@@ -112,7 +112,7 @@ resource "azurerm_kubernetes_cluster_node_pool" "infracipool" {
   orchestrator_version  = local.kubernetes_versions["privatek8s"]
   kubernetes_cluster_id = azurerm_kubernetes_cluster.privatek8s.id
   enable_auto_scaling   = true
-  min_count             = 1
+  min_count             = 0
   max_count             = 20
   zones                 = [3]
   vnet_subnet_id        = data.azurerm_subnet.privatek8s_tier.id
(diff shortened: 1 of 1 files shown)


timja (approved):

privatek8s.tf
```diff
@@ -112,7 +112,7 @@ resource "azurerm_kubernetes_cluster_node_pool" "infracipool" {
   orchestrator_version  = local.kubernetes_versions["privatek8s"]
   kubernetes_cluster_id = azurerm_kubernetes_cluster.privatek8s.id
   enable_auto_scaling   = true
-  min_count             = 0
+  min_count             = 1
```
smerle33, line 115: I would stay at 0 for this nodepool for now as we never encounter the autoscaler problem and would save 33$/month for now
```suggestion
  min_count             = 0
```

timja: this is the case we hit it on, so you may hit it

dduportal (dismissed):

privatek8s.tf
smerle33: we plan to migrate all the agent workload to arm64 so we plan to let the 0 minimum for the non arm64 nodes

timja (approved):

PR #623, opened 2024-02-22
feat(jenkins.io): new premium storage and infra.ci service principal writer

lemeurherve:
This PR adds a new Premium storage account and file share for jenkins.io, and a service principal writer allowing infra.ci.jenkins.io to use azcopy and short lived SAS token to interact with it.

Looking at existing jenkins.io storage account, we noticed that most of its $70/month cost came from transactions.
Using a Premium storage account instead with no transaction cost should allow us to decrease the monthly charge.

Compared to the existing "prodjenkinsio" storage account, this PR sets the new storage account replication type from "GRS" to "LRS", sufficient for our use case.

Ref:

### DIFF
--- a/infra.ci.jenkins.io.tf
+++ b/infra.ci.jenkins.io.tf
@@ -98,6 +98,26 @@ output "infra_ci_jenkins_io_fileshare_serviceprincipal_writer_sp_password" {
   value     = module.infra_ci_jenkins_io_fileshare_serviceprincipal_writer.fileshare_serviceprincipal_writer_sp_password
 }
 
+# Required to allow azcopy sync of jenkins.io File Share
+module "infraci_jenkinsio_fileshare_serviceprincipal_writer" {
+  source = "./.shared-tools/terraform/modules/azure-jenkinsinfra-fileshare-serviceprincipal-writer"
+
+  service_fqdn                   = "infraci-jenkinsio-fileshare_serviceprincipal_writer"
+  active_directory_owners        = [data.azuread_service_principal.terraform_production.id]
+  active_directory_url           = "https://github.com/jenkins-infra/azure"
+  service_principal_end_date     = "2024-07-23T00:00:00Z"
+  file_share_resource_manager_id = azurerm_storage_share.jenkins_io.resource_manager_id
+  storage_account_id             = azurerm_storage_account.jenkins_io.id
+  default_tags                   = local.default_tags
+}
+output "infraci_jenkinsio_fileshare_serviceprincipal_writer_application_client_id" {
+  value = module.infraci_jenkinsio_fileshare_serviceprincipal_writer.fileshare_serviceprincipal_writer_application_client_id
+}
+output "infraci_jenkinsio_fileshare_serviceprincipal_writer_password" {
+  sensitive = true
+  value     = module.infraci_jenkinsio_fileshare_serviceprincipal_writer.fileshare_serviceprincipal_writer_password
+}
+
 locals {
   infra_ci_jenkins_io_fqdn                        = "infra.ci.jenkins.io"
   infra_ci_jenkins_io_service_short_name          = trimprefix(trimprefix(local.infra_ci_jenkins_io_fqdn, "jenkins.io"), ".")
--- a/jenkins.io.tf
+++ b/jenkins.io.tf
@@ -1,0 +1,20 @@
+resource "azurerm_resource_group" "jenkins_io" {
+  name     = "jenkinsio"
+  location = var.location
+}
+
+resource "azurerm_storage_account" "jenkins_io" {
+  name                     = "jenkinsio"
+  resource_group_name      = azurerm_resource_group.jenkins_io.name
+  location                 = azurerm_resource_group.jenkins_io.location
+  account_tier             = "Premium"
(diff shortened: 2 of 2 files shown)


jenkins.io.tf
```diff
@@ -0,0 +1,20 @@
...
+  name     = "jenkinsio"
+  location = var.location
+}
+
+resource "azurerm_storage_account" "jenkins_io" {
+  name                     = "jenkinsio"
+  resource_group_name      = azurerm_resource_group.jenkins_io.name
+  location                 = azurerm_resource_group.jenkins_io.location
+  account_tier             = "Premium"
+  account_replication_type = "LRS"
```
dduportal, line 11:
```suggestion
  account_replication_type = "ZRS"
```

As we need to be able to spin up VMs (or AKS nodes) on different availability zones tro reach this file storage.

jenkins.io.tf
```diff
@@ -0,0 +1,20 @@
+resource "azurerm_resource_group" "jenkins_io" {
+  name     = "jenkinsio"
+  location = var.location
+}
+
+resource "azurerm_storage_account" "jenkins_io" {
+  name                     = "jenkinsio"
+  resource_group_name      = azurerm_resource_group.jenkins_io.name
+  location                 = azurerm_resource_group.jenkins_io.location
+  account_tier             = "Premium"
```
dduportal, line 10: There are missing parameters (some are mandatory otherwise the creation won't succeed despite the terraform plan): see https://github.com/jenkins-infra/azure/blob/5fc5c80a3a97901ad5c99041b375419c0db9dd45/get.jenkins.io.tf#L13-L38

lemeurherve: Thanks, adding them

lemeurherve:
```
Plan: 9 to add, 0 to change, 0 to destroy.

Changes to Outputs:
  + infraci_jenkinsio_fileshare_serviceprincipal_writer_application_client_id   = (known after apply)
  + infraci_jenkinsio_fileshare_serviceprincipal_writer_password                = (sensitive value)
```

LGTM.

No idea yet why updatecli failed its third build on a removed manifest, restarted a new job that passed.

jenkins.io.tf
```diff
@@ -0,0 +1,24 @@
...
+resource "azurerm_storage_account" "jenkins_io" {
+  name                              = "jenkinsio"
+  resource_group_name               = azurerm_resource_group.jenkins_io.name
+  location                          = azurerm_resource_group.jenkins_io.location
+  account_tier                      = "Premium"
+  account_kind                      = "FileStorage"
+  access_tier                       = "Hot"
+  account_replication_type          = "ZRS"
+  min_tls_version                   = "TLS1_2" # default value, needed for tfsec
+  infrastructure_encryption_enabled = true
```
dduportal, line 15: Missing the network rules as well: we don't want this file storage to be available other than through the jenkinsio webservice running in `publick8s` (or inside the trusted and private networks)

dduportal (changes requested):
We have to add network_rules (same as get.jenkins.io)

jenkins.io.tf
```diff
@@ -0,0 +1,41 @@
...
+  network_rules {
+    default_action = "Deny"
+    ip_rules = flatten(
+      concat(
+        [for key, value in module.jenkins_infra_shared_data.admin_public_ips : value],
+      )
+    )
+    virtual_network_subnet_ids = [
+      data.azurerm_subnet.publick8s_tier.id,
+      data.azurerm_subnet.privatek8s_tier.id,
```
dduportal, line 27: Do we write anything on this SAS from infra.ci.jenkins.io or release.ci.jenkins.io?

If yes you can dismiss this suggestion. Otherwise let's remove this line as per the least principle privilege

```suggestion
```

infra.ci.jenkins.io.tf
```diff
@@ -98,6 +98,26 @@ output "infra_ci_jenkins_io_fileshare_serviceprincipal_writer_sp_password" {
...
+  storage_account_id             = azurerm_storage_account.jenkins_io.id
+  default_tags                   = local.default_tags
+}
+output "infraci_jenkinsio_fileshare_serviceprincipal_writer_application_client_id" {
+  value = module.infraci_jenkinsio_fileshare_serviceprincipal_writer.fileshare_serviceprincipal_writer_application_client_id
+}
+output "infraci_jenkinsio_fileshare_serviceprincipal_writer_password" {
+  sensitive = true
+  value     = module.infraci_jenkinsio_fileshare_serviceprincipal_writer.fileshare_serviceprincipal_writer_password
+}
```
dduportal, line 119: Sorry, just caught the issue here but: this should be associated to trusted.ci SP, not infra.ci's

dduportal (changes requested):
Going in a good direction 👍 

- 1 blocker (trusted, not infra SP)
- 1 nit (no need to access from infra.ci vnet)

infra.ci.jenkins.io.tf
lemeurherve: 🤦

dduportal (approved):
`Plan: 9 to add, 0 to change, 0 to destroy.` 🚀

PR #626, opened 2024-02-22
fixup(jenkins.io): correct storage account for `jenkins-io` file share

lemeurherve:
This PR sets the correct storage account for `jenkins-io` file share.

Fixup of:
- #623


dduportal (approved):
ooopsie

PR #633, opened 2024-03-04
feat(privatek8s): adding a large arm64 nodepool

smerle33:
as per https://github.com/jenkins-infra/jenkins.io/pull/7027/files#r1511135932
and https://github.com/jenkins-infra/kubernetes-management/pull/5035/files

need a bigger node for ARM64 for large pods

(we may need to add a taint and toleration to reserve those nodes for only large pods)

Prices for those large nodes are exactly double of the small ones (264,26$/month)

### DIFF
--- a/privatek8s.tf
+++ b/privatek8s.tf
@@ -169,7 +169,7 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
   tags = local.default_tags
 }
 
-resource "azurerm_kubernetes_cluster_node_pool" "infracilgarm64" {
+resource "azurerm_kubernetes_cluster_node_pool" "infracilargearm64" {
   name                  = "arm64large"
   vm_size               = "Standard_D8pds_v5" # 8 vCPU, 32 GB RAM, local disk: 300 GB and 38000 IOPS
   os_disk_type          = "Ephemeral"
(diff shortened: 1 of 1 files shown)


privatek8s.tf
```diff
@@ -169,6 +169,39 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
...
 
+resource "azurerm_kubernetes_cluster_node_pool" "infracilgarm64" {
+  name                  = "arm64large"
+  vm_size               = "Standard_D8pds_v5" # 8 vCPU, 32 GB RAM, local disk: 300 GB and 38000 IOPS
+  os_disk_type          = "Ephemeral"
+  os_disk_size_gb       = 300 # Ref. Cache storage size at https://learn.microsoft.com/en-us/azure/virtual-machines/dpsv5-dpdsv5-series#dpdsv5-series (depends on the instance size)
+  orchestrator_version  = local.kubernetes_versions["privatek8s"]
+  kubernetes_cluster_id = azurerm_kubernetes_cluster.privatek8s.id
+  enable_auto_scaling   = true
+  min_count             = 1
```
lemeurherve, line 180:
```suggestion
  min_count             = 0
```

privatek8s.tf
```diff
@@ -169,6 +169,39 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
...
+    "kubernetes.azure.com/scalesetpriority=spot:NoSchedule",
+  ]
+  lifecycle {
+    ignore_changes = [node_count]
+  }
+
+  tags = local.default_tags
+}
+
+
```
lemeurherve, line 204:
```suggestion

```

privatek8s.tf
smerle33: this can lead to error with spot instances but seeing our usage I agree that we can take the risk

lemeurherve (approved):

dduportal:
For info: failed to apply and reverted in https://github.com/jenkins-infra/azure/pull/635

PR #645, opened 2024-03-18
chore(azurerm providers): specify subscriptions ID as `locals` and update shared tools

smerle33:
(edited) This PR ensures that terraform ignores any local `az` CLI subscription custom setup and uses raw values instead

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -1,4 +1,8 @@
 locals {
+
+  subscription_main      = "dff2ec18-6a8e-405c-8e45-b7df7465acf0"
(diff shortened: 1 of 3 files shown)


timja (approved):

dduportal (approved):
`No changes. Your infrastructure matches the configuration. 

Note: as seen in team screen share, this PR ensures any custom subscription setup for the local `az` command is ignored by Terraform

Note 2: I'm gonna update the message to a `chore` as the deployment system is up to date and does not fail.

PR #658, opened 2024-04-05
feat(privatek8s): add a nodepool dedicated to jenkins controller

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/3823#issuecomment-2036372624

### DIFF
--- a/privatek8s.tf
+++ b/privatek8s.tf
@@ -169,6 +169,32 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
   tags = local.default_tags
 }
 
+
+resource "azurerm_kubernetes_cluster_node_pool" "infracictrlarm64" {
+  name                  = "arm64ctrl"
+  vm_size               = "Standard_D4pds_v5" # 4 vCPU, 16 GB RAM, local disk: 150 GB and 19000 IOPS
+  os_disk_type          = "Ephemeral"
+  os_disk_size_gb       = 150 # Ref. Cache storage size at https://learn.microsoft.com/en-us/azure/virtual-machines/dpsv5-dpdsv5-series#dpdsv5-series (depends on the instance size)
+  orchestrator_version  = local.kubernetes_versions["privatek8s"]
+  kubernetes_cluster_id = azurerm_kubernetes_cluster.privatek8s.id
+  enable_auto_scaling   = true
+  min_count             = 1
+  max_count             = 3
+  zones                 = [1]
+  vnet_subnet_id        = data.azurerm_subnet.privatek8s_tier.id
+
+  node_taints = [
+    "jenkins=infra.ci.jenkins.io:NoSchedule",
+    "kubernetes.azure.com/scalesetpriority=spot:NoSchedule",
+    "kind=controller"
+  ]
+  lifecycle {
+    ignore_changes = [node_count]
+  }
+
+  tags = local.default_tags
+}
+
 resource "azurerm_kubernetes_cluster_node_pool" "releasepool" {
   name                  = "releasepool"
   vm_size               = "Standard_D8s_v3" # 8 vCPU 32 GiB RAM
--- a/privatek8s.tf
+++ b/privatek8s.tf
@@ -147,7 +147,7 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
   enable_auto_scaling   = true
   min_count             = 1
   max_count             = 10
-  zones                 = [1] # zone 1 is dedicated to ARM64 by azure
+  zones                 = [1] # Linux arm64 VMs are only available in the Zone 1 in this region (undocumented by Azure)
   vnet_subnet_id        = data.azurerm_subnet.privatek8s_tier.id
 
   # Spot instances
@@ -185,7 +185,6 @@ resource "azurerm_kubernetes_cluster_node_pool" "infracictrlarm64" {
 
   node_taints = [
     "jenkins=infra.ci.jenkins.io:NoSchedule",
-    "kubernetes.azure.com/scalesetpriority=spot:NoSchedule",
     "kind=controller"
   ]
   lifecycle {
--- a/privatek8s.tf
+++ b/privatek8s.tf
@@ -183,8 +183,8 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
 }
 
 # nodepool dedicated for the infra.ci.jenkins.io controller
-resource "azurerm_kubernetes_cluster_node_pool" "infracictrlarm64" {
-  name                  = "arm64infracictrl"
+resource "azurerm_kubernetes_cluster_node_pool" "infraci_controller" {
+  name                  = "infracictrl"
   vm_size               = "Standard_D4pds_v5" # 4 vCPU, 16 GB RAM, local disk: 150 GB and 19000 IOPS
   os_disk_type          = "Ephemeral"
   os_disk_size_gb       = 150 # Ref. Cache storage size at https://learn.microsoft.com/en-us/azure/virtual-machines/dpsv5-dpdsv5-series#dpdsv5-series (depends on the instance size)
@@ -192,13 +192,13 @@ resource "azurerm_kubernetes_cluster_node_pool" "infracictrlarm64" {
   kubernetes_cluster_id = azurerm_kubernetes_cluster.privatek8s.id
   enable_auto_scaling   = true
   min_count             = 1
-  max_count             = 3
-  zones                 = [1] # zone 1 is dedicated to ARM64 by azure
+  max_count             = 2
+  zones                 = [1] # Linux arm64 VMs are only available in the Zone 1 in this region (undocumented by Azure)
   vnet_subnet_id        = data.azurerm_subnet.privatek8s_infra_ci_controller_tier.id
 
   node_taints = [
     "jenkins=infra.ci.jenkins.io:NoSchedule",
-    "kind=infra.ci.jenkins.io"
+    "jenkins-component=controller:NoSchedule"
   ]
   lifecycle {
     ignore_changes = [node_count]
@@ -208,8 +208,8 @@ resource "azurerm_kubernetes_cluster_node_pool" "infracictrlarm64" {
 }
 
 # nodepool dedicated for the release.ci.jenkins.io controller
-resource "azurerm_kubernetes_cluster_node_pool" "releasecictrlarm64" {
-  name                  = "arm64releasecictrl"
+resource "azurerm_kubernetes_cluster_node_pool" "releaseci_controller" {
+  name                  = "releacictrl"
   vm_size               = "Standard_D4pds_v5" # 4 vCPU, 16 GB RAM, local disk: 150 GB and 19000 IOPS
   os_disk_type          = "Ephemeral"
   os_disk_size_gb       = 150 # Ref. Cache storage size at https://learn.microsoft.com/en-us/azure/virtual-machines/dpsv5-dpdsv5-series#dpdsv5-series (depends on the instance size)
@@ -217,13 +217,13 @@ resource "azurerm_kubernetes_cluster_node_pool" "releasecictrlarm64" {
   kubernetes_cluster_id = azurerm_kubernetes_cluster.privatek8s.id
   enable_auto_scaling   = true
   min_count             = 1
-  max_count             = 3
-  zones                 = [1] # zone 1 is dedicated to ARM64 by azure
+  max_count             = 2
+  zones                 = [1] # Linux arm64 VMs are only available in the Zone 1 in this region (undocumented by Azure)
   vnet_subnet_id        = data.azurerm_subnet.privatek8s_release_ci_controller_tier.id
 
   node_taints = [
     "jenkins=release.ci.jenkins.io:NoSchedule",
-    "kind=release.ci.jenkins.io"
+    "jenkins-component=controller:NoSchedule"
   ]
   lifecycle {
     ignore_changes = [node_count]
--- a/privatek8s.tf
+++ b/privatek8s.tf
@@ -20,6 +20,19 @@ data "azurerm_subnet" "privatek8s_release_tier" {
   virtual_network_name = data.azurerm_virtual_network.private.name
 }
 
+data "azurerm_subnet" "privatek8s_infra_ci_controller_tier" {
+  name                 = "privatek8s-infraci-ctrl-tier"
+  resource_group_name  = data.azurerm_resource_group.private.name
+  virtual_network_name = data.azurerm_virtual_network.private.name
+}
+
+data "azurerm_subnet" "privatek8s_release_ci_controller_tier" {
+  name                 = "privatek8s-releaseci-ctrl-tier"
+  resource_group_name  = data.azurerm_resource_group.private.name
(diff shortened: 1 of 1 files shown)


privatek8s.tf
```diff
@@ -169,6 +169,32 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
   tags = local.default_tags
 }
 
+
+resource "azurerm_kubernetes_cluster_node_pool" "infracictrlarm64" {
+  name                  = "arm64ctrl"
+  vm_size               = "Standard_D4pds_v5" # 4 vCPU, 16 GB RAM, local disk: 150 GB and 19000 IOPS
```
dduportal, line 175: - ✅ infra.ci (2 vCPUS / 8 Go): https://github.com/jenkins-infra/kubernetes-management/blob/24eb85de3ba3720d5638011ed8861d38a3a98dc7/config/jenkins_infra.ci.jenkins.io.yaml#L29-L35
- ✅ release.ci (2 vCPU / 4 Go): https://github.com/jenkins-infra/kubernetes-management/blob/24eb85de3ba3720d5638011ed8861d38a3a98dc7/config/jenkins_release.ci.jenkins.io.yaml#L35-L41

privatek8s.tf
```diff
@@ -169,6 +169,32 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
...
+  min_count             = 1
+  max_count             = 3
+  zones                 = [1] # zone 1 is dedicated to ARM64 by azure
+  vnet_subnet_id        = data.azurerm_subnet.privatek8s_tier.id
+
+  node_taints = [
+    "jenkins=infra.ci.jenkins.io:NoSchedule",
+    "kubernetes.azure.com/scalesetpriority=spot:NoSchedule",
+    "kind=controller"
+  ]
```
dduportal, line 190: We need to rethink the taints/toleration:

- https://github.com/jenkins-infra/kubernetes-management/blob/24eb85de3ba3720d5638011ed8861d38a3a98dc7/config/jenkins_release.ci.jenkins.io.yaml#L228-L236 has a toleration for a `jenkins=release.ci.jenkins.io:NoSchedule` taint. The current `linuxpool` does not have any taint that is why these tolerations are working today
- I see `kind=controller`. But what about the 3 other "Jenkins services" currently running on the `linuxnodepool` (github-ops-comment, rss2twitter and the ircbot) ? Do you plan to schedule them on a distinct nodepool in the loing term (once we will be running full arm64)? Same question for the "side" services such as cert-manager, datadog cluster-agent, nginx ingresses ?

privatek8s.tf
```diff
@@ -169,6 +169,32 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
...
+  kubernetes_cluster_id = azurerm_kubernetes_cluster.privatek8s.id
+  enable_auto_scaling   = true
+  min_count             = 1
+  max_count             = 3
+  zones                 = [1] # zone 1 is dedicated to ARM64 by azure
+  vnet_subnet_id        = data.azurerm_subnet.privatek8s_tier.id
+
+  node_taints = [
+    "jenkins=infra.ci.jenkins.io:NoSchedule",
+    "kubernetes.azure.com/scalesetpriority=spot:NoSchedule",
```
dduportal, line 188:
```suggestion
```

We're not using spot instance ;)

privatek8s.tf
```diff
@@ -147,7 +147,7 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
   enable_auto_scaling   = true
   min_count             = 1
   max_count             = 10
-  zones                 = [1]
+  zones                 = [1] # zone 1 is dedicated to ARM64 by azure
```
dduportal, line 150:
```suggestion
  zones                 = [1] # Linux arm64 VMs are only available in the Zone 1 in this region (undocumented by Azure)
```

dduportal (changes requested):
Nice work!

- 1 nitpick on a comment
- The taints are blocking though: need to clean up spot and to think about the next steps in advance:
  - Do we want to only have the controller on their (new) nodepool here and run other services (non agents) in another? Or do we want to mutualize (might need costs evaluation in a scenario full arm64 withotu agents)

privatek8s.tf
smerle33: you mean that we should have a taint not specific for controller but for "no agent pool" ? to be able to schedule all the others services on it ?

dduportal: > you mean that we should have a taint not specific for controller but for "no agent pool" ? to be able to schedule all the others services on it ?

I mean you have to define the expected goal before implementing with taints/tolerations (e.g. what target architecture are you envisioning?):

- Distinct node pools for controllers and for other "non agent" services (bots, technical services, etc.)?
- Or an unified node pool for all?

=> The sizing (instance size and min/max node) and taints/toleration will be different in both cases.
I believe checking the current sizing, resource usage and costs should be the arguments to decide

privatek8s.tf
```diff
@@ -169,6 +182,55 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
   tags = local.default_tags
 }
 
+# nodepool dedicated for the infra.ci.jenkins.io controller
+resource "azurerm_kubernetes_cluster_node_pool" "infracictrlarm64" {
+  name                  = "arm64infracictrl"
```
dduportal, line 187: As per https://learn.microsoft.com/en-us/azure/aks/create-node-pools#limitations, the Linux node pool names are limited to 12 chars max. As @smerle33 reminded me, the Azure API does not validate the name during the "terraform plan" but only at the "terraform apply".

The `arm64infracictrl` is 16 characters which will fail once the PR is merged.

As such, I propose the following with 11 characters (room to add a suffix number if needed):

```suggestion
resource "azurerm_kubernetes_cluster_node_pool" "infraci_controller" {
  name                  = "infracictrl"
```

- Terraform resource identifier has no character limitation such as Azure API: we can use readable names
- Removing `arm64` from both Terraform Resource identifier and the node pool name: we only indicate the node pool role (e.g. "hosting infra.ci controller" or "hosting release.ci controller")
- Keeping the `ctrl` (see above)

WDYT?

privatek8s.tf
```diff
@@ -169,6 +182,55 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
...
+  lifecycle {
+    ignore_changes = [node_count]
+  }
+
+  tags = local.default_tags
+}
+
+# nodepool dedicated for the release.ci.jenkins.io controller
+resource "azurerm_kubernetes_cluster_node_pool" "releasecictrlarm64" {
+  name                  = "arm64releasecictrl"
```
dduportal, line 212: As per https://learn.microsoft.com/en-us/azure/aks/create-node-pools#limitations, the Linux node pool names are limited to 12 chars max. As @smerle33 reminded me, the Azure API does not validate the name during the "terraform plan" but only at the "terraform apply".

The `arm64releasecictrl` is 18 characters which will fail once the PR is merged.

As such, I propose the following with 11 characters (room to add a suffix number if needed):

```suggestion
resource "azurerm_kubernetes_cluster_node_pool" "releaseci_controller" {
  name                  = "releacictrl"
```

- Terraform resource identifier has no character limitation such as Azure API: we can use readable names
- Removing `arm64` from both Terraform Resource identifier and the node pool name: we only indicate the node pool role (e.g. "hosting infra.ci controller" or "hosting release.ci controller")
- Keeping the `ctrl` (see above)

WDYT?

privatek8s.tf
```diff
@@ -169,6 +182,55 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
...
+  kubernetes_cluster_id = azurerm_kubernetes_cluster.privatek8s.id
+  enable_auto_scaling   = true
+  min_count             = 1
+  max_count             = 3
+  zones                 = [1] # zone 1 is dedicated to ARM64 by azure
+  vnet_subnet_id        = data.azurerm_subnet.privatek8s_infra_ci_controller_tier.id
+
+  node_taints = [
+    "jenkins=infra.ci.jenkins.io:NoSchedule",
+    "kind=infra.ci.jenkins.io"
```
dduportal, line 201:
```suggestion
    "jenkins=infra.ci.jenkins.io:NoSchedule",
    "jenkins-component=controller:NoSchedule"
```

- As per https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/, the taint must have a 3rd elements which is the "Taint effect": adding `: NoSchedule` at the end
- Using `jenkins-component` instead of `kind`: the latter is too generic to have a clear meaning. I propose `jenkins-component` as we keep using "agent / non agent" wording: if we stick to "controller / agent" official wording that will help to make thing clear in the context of a "Jenkins" installation

WDYT?

privatek8s.tf
```diff
@@ -169,6 +182,55 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
...
+  enable_auto_scaling   = true
+  min_count             = 1
+  max_count             = 3
+  zones                 = [1] # zone 1 is dedicated to ARM64 by azure
+  vnet_subnet_id        = data.azurerm_subnet.privatek8s_release_ci_controller_tier.id
+
+  node_taints = [
+    "jenkins=release.ci.jenkins.io:NoSchedule",
+    "kind=release.ci.jenkins.io"
+  ]
```
dduportal, line 227:
```suggestion
  node_taints = [
    "jenkins=release.ci.jenkins.io:NoSchedule",
    "jenkins-component=controller:NoSchedule"
  ]
```

- As per https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/, the taint must have a 3rd elements which is the "Taint effect": adding `: NoSchedule` at the end
- Using `jenkins-component` instead of `kind`: the latter is too generic to have a clear meaning. I propose `jenkins-component` as we keep using "agent / non agent" wording: if we stick to "controller / agent" official wording that will help to make thing clear in the context of a "Jenkins" installation

WDYT?

privatek8s.tf
```diff
@@ -169,6 +182,55 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
...
+  name                  = "arm64infracictrl"
+  vm_size               = "Standard_D4pds_v5" # 4 vCPU, 16 GB RAM, local disk: 150 GB and 19000 IOPS
+  os_disk_type          = "Ephemeral"
+  os_disk_size_gb       = 150 # Ref. Cache storage size at https://learn.microsoft.com/en-us/azure/virtual-machines/dpsv5-dpdsv5-series#dpdsv5-series (depends on the instance size)
+  orchestrator_version  = local.kubernetes_versions["privatek8s"]
+  kubernetes_cluster_id = azurerm_kubernetes_cluster.privatek8s.id
+  enable_auto_scaling   = true
+  min_count             = 1
+  max_count             = 3
+  zones                 = [1] # zone 1 is dedicated to ARM64 by azure
```
dduportal, line 196:
```suggestion
  zones                 = [1] # Linux arm64 VMs are only available in the Zone 1 in this region (undocumented by Azure)
```

(nit) 

- The current comment is unclear: Zone 1 can have amd64 machine so the word "dedicated" is not valid
- `arm64` must be lower cased

privatek8s.tf
```diff
@@ -169,6 +182,55 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
...
+  name                  = "arm64releasecictrl"
+  vm_size               = "Standard_D4pds_v5" # 4 vCPU, 16 GB RAM, local disk: 150 GB and 19000 IOPS
+  os_disk_type          = "Ephemeral"
+  os_disk_size_gb       = 150 # Ref. Cache storage size at https://learn.microsoft.com/en-us/azure/virtual-machines/dpsv5-dpdsv5-series#dpdsv5-series (depends on the instance size)
+  orchestrator_version  = local.kubernetes_versions["privatek8s"]
+  kubernetes_cluster_id = azurerm_kubernetes_cluster.privatek8s.id
+  enable_auto_scaling   = true
+  min_count             = 1
+  max_count             = 3
+  zones                 = [1] # zone 1 is dedicated to ARM64 by azure
```
dduportal, line 221:
```suggestion
  zones                 = [1] # Linux arm64 VMs are only available in the Zone 1 in this region (undocumented by Azure)
```

(nit) 

- The current comment is unclear: Zone 1 can have amd64 machine so the word "dedicated" is not valid
- `arm64` must be lower cased

privatek8s.tf
```diff
@@ -169,6 +182,55 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
...
+resource "azurerm_kubernetes_cluster_node_pool" "infracictrlarm64" {
+  name                  = "arm64infracictrl"
+  vm_size               = "Standard_D4pds_v5" # 4 vCPU, 16 GB RAM, local disk: 150 GB and 19000 IOPS
+  os_disk_type          = "Ephemeral"
+  os_disk_size_gb       = 150 # Ref. Cache storage size at https://learn.microsoft.com/en-us/azure/virtual-machines/dpsv5-dpdsv5-series#dpdsv5-series (depends on the instance size)
+  orchestrator_version  = local.kubernetes_versions["privatek8s"]
+  kubernetes_cluster_id = azurerm_kubernetes_cluster.privatek8s.id
+  enable_auto_scaling   = true
+  min_count             = 1
+  max_count             = 3
```
dduportal, line 195:
```suggestion
  max_count             = 2
```

=> 1 controller per node means no need to surge up to 3 nodes

privatek8s.tf
```diff
@@ -169,6 +182,55 @@ resource "azurerm_kubernetes_cluster_node_pool" "infraciarm64" {
...
+resource "azurerm_kubernetes_cluster_node_pool" "releasecictrlarm64" {
+  name                  = "arm64releasecictrl"
+  vm_size               = "Standard_D4pds_v5" # 4 vCPU, 16 GB RAM, local disk: 150 GB and 19000 IOPS
+  os_disk_type          = "Ephemeral"
+  os_disk_size_gb       = 150 # Ref. Cache storage size at https://learn.microsoft.com/en-us/azure/virtual-machines/dpsv5-dpdsv5-series#dpdsv5-series (depends on the instance size)
+  orchestrator_version  = local.kubernetes_versions["privatek8s"]
+  kubernetes_cluster_id = azurerm_kubernetes_cluster.privatek8s.id
+  enable_auto_scaling   = true
+  min_count             = 1
+  max_count             = 3
```
dduportal, line 220:
```suggestion
  max_count             = 2
```

=> 1 controller per node means no need to surge up to 3 nodes

dduportal (changes requested):
Nice work! 

A few feedbacks:

- blocker: the node pools names are too big. I've added suggestions.
- blocker: the taints are invalids. I've added a proposal to make them valid + to help us making the "node pool roles" clearer
- not blocking: let's limit the max surge to 2 nodes as we plan to have 1 controller per node
- nitpicking: comment suggestions

Also, Could you, once you have a clear view on this PR, update the PR body to give details about:

- Choice of sizing (why a smaller node - explaining the "1 controller per node" pattern, but also the billing expected impact comapred to the current amd64 sizing)
- A reminder of why using a subnet per node pool
- A note explaining why we plan to use 1 node pool per controller, and not for other services

PR #662, opened 2024-04-12
chore(privatek8s): change from ubuntu to azure linux the os for the nodes

smerle33:
as per https://learn.microsoft.com/en-us/azure/azure-linux/intro-azure-linux#azure-linux-container-host-key-benefits
AzureLinux OS is lighter and more prone to fast security update.

Lets take advantage to those 2 not yet used nodepool to switch from ubuntu to azurelinux


timja (approved):
Interesting, I haven't tried it myself although been aware of it for awhile, good to see its matured. Hopefully it works well

dduportal (approved):

PR #672, opened 2024-04-22
feat(privatek8s/publick8s) add a standard ZRS storage class

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4044

add a new storage class for ZRS in standard mode to be used for weekly.ci on the `publick8s` and infra.ci/release.ci on the `privatek8s`


privatek8s.tf
```diff
@@ -339,7 +339,7 @@ resource "kubernetes_storage_class" "azurefile_csi_premium_retain" {
   provider      = kubernetes.privatek8s
 }
 
-resource "kubernetes_storage_class" "managed_csi_premium_ZRS_retain_public_private" {
+resource "kubernetes_storage_class" "managed_csi_premium_ZRS_retain_private" {
```
dduportal, line 342: This change is legit (e.g. using a correct "internal" name for the terraform resource) but it is a refactoring and requires a `moved` block (see https://developer.hashicorp.com/terraform/language/modules/develop/refactoring) to avoid deleting / re-creating the resource in production

dduportal (approved):
- `Plan: 2 to add, 0 to change, 0 to destroy.`
- The mapping providers <-> storage class is good 👍 

LGTM!


Note: the storage class setup looks like an opportunity to factorize in the long term future to avoid risks of confusions with the providers.

dduportal:
@smerle33 you can merge whenever you are ready!

dduportal (approved):

PR #676, opened 2024-04-25
chore(updates.jenkins.io): add a dedicated File Share for httpd service

lemeurherve:
This PR adds a dedicated File Share for httpd.

This will allow us to mount the .htaccess file generated by https://github.com/jenkins-infra/update-center2/blob/master/site/generate-htaccess.sh and completed with https://github.com/jenkins-infra/update-center2/pull/776 only as volume for the `httpd` updates.jenkins.io service, and every other files except .htaccess as volume for the `mirrorbits` updates.jenkins.io service.

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2075477428


lemeurherve:
> Plan: 1 to add, 0 to change, 0 to destroy.

PR #681, opened 2024-05-02
feat: add `docs.jenkins.io` storage account and file share

lemeurherve:
This PR adds the storage account and file share that will be used by the nginx-website release on publick8s cluster to serve on docs.origin.jenkins.io the content generated from https://github.com/jenkins-infra/docs.jenkins.io/.

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/3885#issuecomment-2090052823

### DIFF
--- a/docs.jenkins.io.tf
+++ b/docs.jenkins.io.tf
@@ -1,0 +1,37 @@
+resource "azurerm_resource_group" "docs_jenkins_io" {
+  name     = "docs-jenkins-io"
+  location = var.location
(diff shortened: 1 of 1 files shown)


dduportal (approved):
- `Plan: 3 to add, 0 to change, 0 to destroy.` ✅ 
- We're using a Standard account storage. It means we'll have to check, in a few months, the cost incurred by this service as we're billed by transactions (for each mount, read but also write)

PR #682, opened 2024-05-02
feat!(publick8s/updates.jenkins.io) switch Storage Account to Premium to decrease costs (BREAKING)

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2090967319

This PR switches to a `Premium` storage account kind for updates.jenkins.io to remove costs related to file storage transactions. The storage costs will increase from $0.09 to $0.50 which is ... affordable... to say the least (compared to the expected decrease of ~$320 transaction cost).

Note:
- The update center job is not trying to update storage
- Services azure.updates.jenkins.io and mirrors.updates.jenkins.io have been uninstalled

PR #683, opened 2024-05-02
fix(updates.jenkins.io) set FS to minimum size for each

dduportal:
Fixup of #682 which failed to deploy (HTTP/400 error) when creating the file shares.

As per https://learn.microsoft.com/en-us/azure/storage/files/understanding-billing#provisioning-method, the minimum size for a file share is 100 Gb.

The cost evaluation in https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2090967319 need to be re-evaluated:

- Transactions cost from ~$380 to $0 monthly
- Storage cost from $0.09 monthly to $32 (2 x 100 x 0,16).


dduportal:
Self-merging

PR #698, opened 2024-05-13
fix(ci.jenkins.io-agents-1) remove AKS subnet NSG to avoid kubelet and pod issues

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3954

After a pairing session with @smerle33 where we discovered that the applications running in the AKS cluster's `kube-system` namespace where failing due to timeouts, we realized that the NSG in charge fo setting up rules for inbound agents is not shaped to handle an AKS cluster as it blocks technical requests within the cluster:

- Pods (`10.100.0.0/14`) to internal DNS (`10.0.0.0`) server
- Kubelet (unknown CIDR) to pods (`10.100.0.0/14`)
- Pods (`10.100.0.0/14`) to Kubernetes services
- Etc.

This PR removes the NSG (for now) until we have deployed the cluster

### DIFF
--- a/ci.jenkins.io.tf
+++ b/ci.jenkins.io.tf
@@ -78,27 +78,7 @@ module "ci_jenkins_io_aci_agents_sponsorship" {
   controller_service_principal_id = module.ci_jenkins_io_sponsorship.controller_service_principal_id
 }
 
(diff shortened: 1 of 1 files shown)


dduportal (commented):
- Rules manually deleted so only `Plan: 0 to add, 0 to change, 2 to destroy.`

PR #701, opened 2024-05-16
fix(ci.jenkins.io-kubernetes-agents) use correct labels for the BOM node pool

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3954#issuecomment-2114950444

This PR fixes the label `role` used by the BOM node pool on the `ci.jenkins.io-kubernetes-agents-1` cluster to ensure we can use a specific `nodeSelector` for the BOM container agents


dduportal (commented):

PR #700, opened 2024-05-15
feat(weekly.ci) PVC for weekly.ci.jio

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4044

create a dedicated PVC for weekly.ci that will create a PV and Disk

PR #713, opened 2024-06-04
feat(weekly.ci) add role definition and assignment for new disk access

smerle33:
following #700 and as per https://github.com/jenkins-infra/helpdesk/issues/4044
this PR will allow a pod to access the new PV/PVC/DISK and handle the data migration.

this PR was tested locally, it failed when using both ID and name for the role assignment/role definition, it failed with only the id with `Error: authorization.RoleAssignmentsClient#Create: Failure responding to request: StatusCode=400 -- Original Error: autorest/azure: Service returned an error. Status=400 Code="InvalidRoleDefinitionId" Message="The role definition ID '5a60a9a4-88e7-d46c-6311-6aa2bba1581c|' is not valid."`

but succeeded with only the `role_definition_name` ...

### DIFF
--- a/weekly.ci.jenkins.io.tf
+++ b/weekly.ci.jenkins.io.tf
@@ -53,3 +53,18 @@ resource "kubernetes_persistent_volume_claim" "jenkins_weekly_data" {
     }
   }
 }
+
+# Required to allow the weekly controller to read the disk
+resource "azurerm_role_definition" "weekly_ci_jenkins_io_controller_disk_reader" {
+  name     = "ReadWeeklyCIDisk"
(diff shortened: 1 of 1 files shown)


smerle33:
`Apply complete! Resources: 2 added, 0 changed, 0 destroyed.`
and then
`Destroy complete! Resources: 2 destroyed.`

dduportal (dismissed):
We might need other permissions than `disk/read`: LGTM but worth cehcking if there is a doc. on azure

dduportal (approved):

LGTM!

PR #714, opened 2024-06-04
feat(ci.jenkins.io-agents1) set AKS SKU to `Standard` instead of `Free`

dduportal:
When a cluster has 10+ nodes, the following recommendation pops in Azure portal:

Capture d’écran 2024-06-06 à 14 09 43

This PR changes the SKU on this cluster to ensure better scheduling performances when BOM builds are running.
The cost can be ignored: it is in the subscription paid by sponsored credits


timja (approved):
If we aren't paying for it then no brainer.

smerle33 (approved):
🚀

lemeurherve (approved):

PR #727, opened 2024-06-14
chore(AKS clusters) explicit upgrade_settings.max_surge to default Azure value for all nodepools to avoid permanent update

smerle33:
as per 
- https://github.com/hashicorp/terraform-provider-azurerm/issues/24020

we are having the same repetitive update on azure, so setting the default should avoir permanent update override


dduportal:
nit: could you update the title to have a valid scope? There are no file or component named `upgrade_settings` in the repository.
Proposal: `chore(AKS clusters) explicit `upgrade_settings.max_surge` to default Azure value for all nodepools to avoid permanent update`

dduportal (approved):
`Plan: 0 to add, 18 to change, 0 to destroy.` LGTM (my nit comment about title is not blocking and can be changed after merging)

PR #729, opened 2024-06-16
feat: add `stats.jenkins.io` storage account and file share

lemeurherve:
This PR adds the storage account and file share that will be used by the nginx-website release on publick8s cluster to serve on new.stats.jenkins.io the content generated from https://github.com/jenkins-infra/stats.jenkins.io/.

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/4132#issuecomment-2167717401

### DIFF
--- a/stats.jenkins.io.tf
+++ b/stats.jenkins.io.tf
@@ -1,0 +1,37 @@
+resource "azurerm_resource_group" "stats_jenkins_io" {
+  name     = "stats-jenkins-io"
+  location = var.location
+  tags     = local.default_tags
+}
+
+resource "azurerm_storage_account" "stats_jenkins_io" {
+  name                      = "statsjenkinsio"
+  resource_group_name       = azurerm_resource_group.stats_jenkins_io.name
+  location                  = azurerm_resource_group.stats_jenkins_io.location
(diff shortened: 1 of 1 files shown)


timja (approved):

stats.jenkins.io.tf
```diff
@@ -0,0 +1,37 @@
...
+  name     = "stats-jenkins-io"
+  location = var.location
+  tags     = local.default_tags
+}
+
+resource "azurerm_storage_account" "stats_jenkins_io" {
+  name                      = "statsjenkinsio"
+  resource_group_name       = azurerm_resource_group.stats_jenkins_io.name
+  location                  = azurerm_resource_group.stats_jenkins_io.location
+  account_tier              = "Standard"
```
dduportal, line 11: Note: if you use a Standard storage account, you'll have to check the cost due to amount of transactions. It usually are the write which are costly: what would be the statistic's update frequency?

I believe we can for it but you have to watch the costs of this resource group in the upcoming (4-6) weeks

dduportal (approved):
LGTM! Please watch the costs in the upcoming weeks if you keep the Standard storage

PR #737, opened 2024-06-20
chore(updatecli): track `infraci_stats_jenkins_io_fileshare_serviceprincipal_writer` end date

lemeurherve:
This PR adds an updatecli manifest to track `infraci_stats_jenkins_io_fileshare_serviceprincipal_writer` end date.

Proposed PR body:

> image

Test:
<details><summary>updatecli diff --values updatecli/values.yaml --config updatecli/updatecli.d/stats.jenkins.io.sp.endate.yaml</summary>

```console
+++++++++++
+ PREPARE +
+++++++++++

Loading Pipeline "updatecli/updatecli.d/stats.jenkins.io.sp.endate.yaml"

SCM repository retrieved: 1


++++++++++++++++++
+ AUTO DISCOVERY +
++++++++++++++++++



++++++++++++
+ PIPELINE +
++++++++++++



##################################################################################
# GENERATE NEW END DATE FOR STATS.JENKINS.IO FILE SHARE SERVICE PRINCIPAL WRITER #
##################################################################################


SOURCES
=======

nextExpiry
----------
The shell 🐚 command "/bin/sh /var/folders/ly/77mv7l9968g068w9_fwzwggw0000gp/T/updatecli/bin/bd1b66faf6e39c527f085eea5b269bba76c890efa7550e4ca84317fcabb6e3f1.sh" ran successfully with the following output:
----
2024-09-18T00:00:00Z
----
✔ shell command executed successfully

currentExpiry
-------------
✔ value "\"2023-09-19T23:00:00Z\"" found for key "$.end_dates.infraci_stats_jenkins_io_fileshare_serviceprincipal_writer" in the yaml file "locals.yaml"


CONDITIONS:
===========

checkIfExpirySoonExpired
------------------------
The shell 🐚 command "/bin/sh /var/folders/ly/77mv7l9968g068w9_fwzwggw0000gp/T/updatecli/bin/56202140e385731a2fc616bc17989ad91eed830fd60ba795da3fdd132df9fd37.sh" ran successfully with the following output:
----
time for update
----
✔ shell condition of type "console/output", passing


TARGETS
========

updateNextExpiry
----------------

**Dry Run enabled**

⚠ - change detected:
        * key "$.end_dates.infraci_stats_jenkins_io_fileshare_serviceprincipal_writer" should be updated from "\"2024-09-19T23:00:00Z\"" to "2024-09-18T00:00:00Z", in file "locals.yaml"


ACTIONS
========


[...]

PR #740, opened 2024-06-21
chore(updatecli): track all File Share service principal writer `end_date`s

lemeurherve:
This PR adds a generic File Share service principal writer end dates manifest.

PR body with stats.jenkins.io as example:

> image

Follow-up of:
- #741
- #737


dduportal (approved):
- Terraform : `No changes. Your infrastructure matches the configuration.` (expected as no TF changes)
- Updatecli: 3 condition not met as expected (like #741 ), and a new end date will be updated (introduced by this PR)


LGTM

PR #743, opened 2024-06-21
fix(updatecli): correct secrets and short next end date in `end-dates` PR body

lemeurherve:
This PR adds concerned secrets to the values to fix the PR body message, and use a short next end date in PR title and commit message.

It also adds the current end date in the PR body.

Ref:
- https://github.com/jenkins-infra/azure/pull/742#issuecomment-2182677940


lemeurherve:
> Note sure why but the updatecli diff fails with `ERROR: ✗ 4 over 14 pipeline failed` :'(

Because the "secret" field doesn't exists on main maybe?

dduportal (approved):

lemeurherve:
> Note sure why but the updatecli diff fails with `ERROR: ✗ 4 over 14 pipeline failed` :'(

This issue comes from `The current end date is set to {{ source "currentEndDate" }}.`

Currently replacing it.

PR #748, opened 2024-06-24
chore: move `trusted.ci.jenkins.io.tf` service principal writer end dates to `updatecli/values.yaml`

lemeurherve:
This PR moves `trusted.ci.jenkins.io.tf` service principal writer end dates to `updatecli/values.yaml`.

A dedicated updatecli manifest will be added in a follow-up pull request so it can be properly validated. (Will adapt the existing manifest at the same time)

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/4148#issuecomment-2185911564


lemeurherve:
> No changes. Your infrastructure matches the configuration.

dduportal (approved):
🥳

PR #754, opened 2024-06-26
chore: rename File Share service principal writers

lemeurherve:
This PR renames existing resources to more descriptive/correct names to avoid mistakes and oversights.

PR #755, opened 2024-06-26
chore: remove service principal id and password outputs and rename application ones

lemeurherve:
This PR removes service principal id and password outputs and rename application ones to be more descriptive about the fact they concern the application client id and the application client password, not the service principal.

It also removes the `moved` blocks introduced in #754 as these renaming are now in terraform state.

Follow-up of:
- https://github.com/jenkins-infra/shared-tools/pull/150
- #754

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/4149


dduportal (approved):
`You can apply this plan to save these new output values to the Terraform
state, without changing any real infrastructure.`


LGTM

PR #759, opened 2024-06-27
chore(infra.ci.jenkins.io) move end date for the Azure client password used by Azure VM agents plugin

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4154

This PR moves the end date definition to the YAML file used by `updatecli` to track dates.

- Local tests shows there are no changes in Terraform plan: let's see what the CI checks tells us
- Bumped the shared tools to ensure we have the latest version all the time
- Naming: I used `azurevms_agents_client_secret` in the YAML file:
  - The prefix `infraci_` does not make sense as this key is under the `infra.ci.jenkins.io` key
  - There are no corresponding environment variable that would fit the other keys naming convention: most probably because we started by this one, and then the resource sets grew over the initial naming.


dduportal:
Closing in favor of #760 as I was mistaken:

The data structure I changed in `values.yaml` is only used by the updatecli manifests used for the file shares SPs (as per the filenames). As such, the templates for PR body won't match at all.

PR #767, opened 2024-07-02
chore(updatecli) simplify manifests using gotemplates by using root scope to get top-level values

dduportal:
This PR simplifies the Go template used in `updatecli` manifests to have less code, which follows the usual Go conventiojns to ensure any contributor with go background will be at ease. It also prevent potential bugs around variables scope re-entrance in `updatecli` and human mistake (overriding accidentally the `github` values).

Go template system (see https://pkg.go.dev/text/template#hdr-Variables) allows accessing the root context (e.g. the values passed to the template engine scoped at the top of the file) through the variable `$`.


lemeurherve (approved):
Thanks!

PR #769, opened 2024-07-02
chore(weekly.ci) make resource dependencies explicit by avoiding locals

dduportal:
We have the following rule of thumb in the Jenkins Infra terraform projects: "If a value is duplicated then factorize it as a local". It's a rule of thumb which has been carefully followed which is good 👍 

However in this particular case (weekly.ci persistent resources), the values duplication factorized into a local has an unexpected side-effect: it hides a dependency chain between resources.

This PR fixes it by making the dependency between resources (PVC -> PV -> disk) explicit.
Positive side-effect: it removes the local so the specific values sides within the scope of the weekly.ci.jenkins.io TF file.

We expect no changes in this PR (for the Terraform plan).

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -62,8 +62,5 @@ locals {
 
   publick8s_compute_zones = [3]
 
-  weekly_ci_disk_size    = 8
-  weekly_ci_access_modes = ["ReadWriteOnce"]
-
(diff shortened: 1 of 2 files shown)


smerle33:
No changes. Your infrastructure matches the configuration.

LGTM

smerle33 (approved):
🚀

PR #771, opened 2024-07-03
feat(infra.ci): create standard ZRS PV/PVC/DISK for infra.ci

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4044

preparing the migration from premium to standard ZRS

PR #773, opened 2024-07-04
fix(infraci): correct the resource group for the disk on the main subscription

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4044#issuecomment-2208414161
following #771


dduportal (approved):
```
Plan: 6 to add, 0 to change, 2 to destroy.
```

PR #774, opened 2024-07-05
cleanup: tune setup of AKS PV disks

dduportal:
- Use proper tags for resource groups
- Reword comment around disk permissions to give the real target (CSI driver, not the controller pods)
- Rename the publick8s static storage class to avoid confusion
- Update the shared tools to latest version


dduportal:
Plan reports:

- tags updated in RGs: `Plan: 0 to add, 3 to change, 0 to destroy.`
- And 1 moved resource

smerle33 (approved):
🚀

PR #780, opened 2024-07-08
chore(updatecli): add current end date to PR titles

lemeurherve:
This PR adds current end date to PR titles.

It also renames the templates as .tpl instead of .yaml, replaces "ExpiryDate" by "EndDate", ignores yamllint "line too long" rule, adds labels and lints the manifests.

### DIFF
--- a/updatecli/updatecli.d/ci.sp.endate.yaml
+++ b/updatecli/updatecli.d/ci.sp.endate.yaml
@@ -1,3 +1,5 @@
+---
+# yamllint disable rule:line-length
 name: "Generate new end date for ci.jenkins.io controller service principal"
 
 scms:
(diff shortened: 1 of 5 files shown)


updatecli/updatecli.d/ci.sp.endate.yaml
```diff
@@ -13,46 +15,52 @@ scms:
...
     scmid: default
 
 actions:
   default:
     kind: github/pullrequest
     scmid: default
     spec:
-      title: Generate a new Azure Service Principal password with expiration date {{ source "nextExpiry" }} on `ci.jenkins.io`
-      description: "Generate a new password for the Azure Service Principal `ci.jenkins.io`. See https://github.com/jenkins-infra/helpdesk/issues/4052#issuecomment-2072867491 for details"
+      title: "Extend Azure Service Principal password on `ci.jenkins.io` (current end date: {{ source \"currentEndDate\" }})"
```
dduportal, line 58: This line breaks the `updatecli` due to wrong interpolation (mix of YAML and GO template):

- Better to use simple quotes for strings
- It allows passing double quotes as litteral char. so the Go template can use it

```suggestion
      title: 'Extend Azure Service Principal password on `ci.jenkins.io` (current end date: {{ source "currentEndDate" }})'
```

dduportal (changes requested):
Almost there. See suggestion

PR #783, opened 2024-07-11
feat: add a test area for Jay onboarding

dduportal:


dduportal:
`Plan: 22 to add, 13 to change, 0 to destroy.` 

- 22 to add as part of this PR
- 13 to change due to my secondary public Ip removal in shared tools

PR #794, opened 2024-08-06
feat(dockerhub-mirror) set up a new dedicated ACR to mirror DockerHub inside the Jenkins Azure infrastructure

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4192

Fixup of https://github.com/jenkins-infra/azure/commit/91cf2dc8eb13120f20add9baddb1a44196f371bd

Reference Azure documentation: https://learn.microsoft.com/en-us/azure/container-registry/container-registry-artifact-cache?pivots=development-environment-azure-portal

This PR introduces an Azure Container Registry set up as a DockerHub mirror using a "Cache Rule" which mirrors `docker.io/*` to `*` (note: it forbids us to use other caching mechanism!).

This registry has the following properties:

- Only available in the "sponsorship" subscription
- Anonymous pull access (constraint due to Docker pull through cache - https://github.com/moby/moby/issues/30880)
- Private network only: since we have anonymous pull policy (see above), then we restrict to only a subset of private networks. It uses ["Azure Private Endpoints"](https://learn.microsoft.com/en-us/azure/private-link/private-endpoint-overview) for this
  - Note: it implies using Private DNS zones linked to networks. These zone might need to be reused in the future for other private link if required

The registry is available for the following (heavy DockerHub users) services (I've only setup the Azure ephemeral VM agents subnets for now) through a combination of (private endpoint with a NIC in the subnet + private DNS zone with automatic records + inbound and outbound NSG rules):
- ci.jenkins.io
- cert.ci.jenkins.io
- trusted.jenkins.io
- infra.jenkins.io

Azure makes it mandatory to log-in on DockerHub for such a mirror system. As such, we use a distinct token stored in an Azure Keyvault which is "Public Images Read Only" associated to the `jenkinsciinfra` organization to avoid the "application" rate limit (e.g. 5k pull / day / IP) and only have the DockerHub anti-abuse system as upper limit (which seems to be a combination of requests and amount of data).

Capture d’écran 2024-08-05 à 16 31 38

----

*Testing and approving*

This PR is expected to have no changes in the plan as it was applied manually:

- End to end testing was done on each controller by:
  - Starting an Azure VM ephemeral agent using a pipeline replay with correct label
    - The pipeline tries to resolve the DNS name `dockerhubmirror.azurecr.io` and should resolve to an IP local to the VM subnet
  - Once the VM is up, checking the connectivity in Azure UI portal (`Network Watcher` -> `Connection troubleshoot`)
    - Source VM is the agent VM, which name is retrieved from build log
    - Destination is `https://dockerhubmirror.azurecr.io`

Capture d’écran 2024-08-06 à 10 42 25

- The bootstrap must be done in 2 `terraform apply` commands as documented, because the ACR component `CredentialSet` is not supported by Terraform yet (see comments in TF code).

### DIFF
--- a/cert.ci.jenkins.io.tf
+++ b/cert.ci.jenkins.io.tf
@@ -117,3 +117,41 @@ resource "azurerm_dns_a_record" "cert_ci_jenkins_io" {
   ttl                 = 60
   records             = [module.cert_ci_jenkins_io.controller_private_ipv4]
 }
+
+## Allow access to/from ACR endpoint
+resource "azurerm_network_security_rule" "allow_out_https_from_cert_ephemeral_agents_to_acr" {
+  provider                = azurerm.jenkins-sponsorship
+  name                    = "allow-out-https-from-ephemeral-agents-to-acr"
+  priority                = 4050
+  direction               = "Outbound"
+  access                  = "Allow"
+  protocol                = "Tcp"
+  source_port_range       = "*"
+  destination_port_range  = "443"
+  source_address_prefixes = data.azurerm_subnet.cert_ci_jenkins_io_sponsorship_ephemeral_agents.address_prefixes
+  destination_address_prefixes = distinct(
+    flatten(
+      [for rs in azurerm_private_endpoint.dockerhub_mirror["certcijenkinsio"].private_dns_zone_configs.*.record_sets : rs.*.ip_addresses]
+    )
+  )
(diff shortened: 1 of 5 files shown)


dduportal:
Merging to fix the Terraform errors on main branch and avoid accidental roll backs.

@timja do not hesitate to review: I'll fix any feedbacks!

PR #803, opened 2024-08-16
feat(publick8s) add a persistent volume for maxmind/geoip

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4240

This PR adds a new PVC statically provisioned based on an Azurefile to ensure a `ReadyMany*` access mode.
The goal is to have a centralized data dir. for the GeoIP database.

- Using a storage account reusable for the whole cluster, hence the naming. Type is Storage v2: means billing per request: only define file shares with low workload!

### DIFF
--- a/publick8s.tf
+++ b/publick8s.tf
@@ -266,6 +266,115 @@ resource "kubernetes_storage_class" "azurefile_csi_premium_retain_public" {
   provider      = kubernetes.publick8s
 }
 
+# Cluster persistent data volumes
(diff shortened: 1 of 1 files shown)


dduportal (commented):
- Checks: `Plan: 6 to add, 0 to change, 0 to destroy.`
- Tested manually (and then cleaned up)

=> only unknown is the main branch in infra.ci (terraform Azure SP)

PR #805, opened 2024-08-20
feat(publick8s) allow cluster to manage private data dtier subnet to allow creating internal LB

dduportal:
While working on https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2294769030,

it appeared that we want the `publick8s` cluster to manage LB in the private data subnet of the **private** network as the private ingress is only expected to be accessed through the private VPN.

It also allows trusted.ci to access it (through the vnet peering trusted -> private).


dduportal (commented):
`Plan: 1 to add, 0 to change, 0 to destroy.` let's roll!

PR #811, opened 2024-08-29
feat(redis): add a public instance of redis for mirrorbits usage

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4259

prepare a P2 instance for REDIS to use for both mirrorbits usage in publicK8S

Plan: 7 to add, 0 to change, 0 to destroy.

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -3,6 +3,10 @@ resource "local_file" "jenkins_infra_data_report" {
     "artifact-caching-proxy.privatelink.azurecr.io" = {
       "service_ip" = tolist(azurerm_private_dns_a_record.artifact_caching_proxy.records)[0],
     },
+    "public_redis" = {
+      "service_hostname" = azurerm_redis_cache.public_redis.hostname,
+      "service_port" = azurerm_redis_cache.public_redis.port,
+    },
   })
   filename = "${path.module}/jenkins-infra-data-reports/azure.json"
 }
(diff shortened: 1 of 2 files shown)


public.redis.tf
```diff
@@ -0,0 +1,61 @@
...
+resource "azurerm_resource_group" "public_redis" {
+  name     = "public-redis"
+  provider = azurerm.jenkins-sponsorship
+  location = var.location
+  tags     = local.default_tags
+}
+
+# Redis database
+resource "azurerm_redis_cache" "public_redis" {
+  name                          = "public-redis"
```
dduportal, line 11:
```suggestion
  name                          = "public-redis"
  provider = azurerm.jenkins-sponsorship
```
=> need to use the sponsored subscription

(dont forget to `terraform fmt -recursive after this suggestion)

dduportal (approved):
LGTM!

Tested during a pair session with a custom `apply` and we verified that:

- Resources where created with no error (ref. 2 different subscriptions => eventual scope constraints)
- Redis instance was created with the expected network properties (no public access, and a private link service defined with a private endpoint)
- Private endpoint is linked to the private DNS zone and the DNS records were inserted by Azure automatically (same as with the registry, ref. https://learn.microsoft.com/en-us/azure/private-link/private-endpoint-dns#databases for namings)
- The `publick8s` cluster has the expected DNS resolution (`dig <service hostname>` does respond the private IP of the private endpoint) and is able to reach the Redis private service through it (`curl -v <redis hostname>:6379` responds "HTTP 0.9  Not Supported")

PR #843, opened 2024-09-20
feat(updates.jenkins.io) add outputs and missing Azure SP to allow using new credentials

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2364054995

This PR adds a missing Azure SP to allow trusted.ci to write data into the new `redirects` UC file share and sets up outputs to ease the credential ZIP script generation.


dduportal:
`Plan: 7 to add, 0 to change, 1 to destroy.` => 6 are Azure AD resources, 1 is the local report.
+ outputs changes

PR #845, opened 2024-09-24
feat(updates.jenkins.io) define statically provisioned PVCs for mirrorbits data and geoipdata

dduportal:
Since https://github.com/jenkins-infra/helm-charts/pull/1339, we can specify existing PVC for mirrorbits usage.

This PR defines statically provisioned (with the new typo-less storage class) volumes for the mirrorbits and geoip datas for (mirrors.)updates.jenkins.io.

It will allows us to have less setup and config in kubernetes management while keeping persistent volume under terraform control

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -9,22 +9,25 @@ resource "local_file" "jenkins_infra_data_report" {
     },
     "updates.jenkins.io" = {
       "content" = {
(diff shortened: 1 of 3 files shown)


dduportal:
`Plan: 5 to add, 0 to change, 1 to destroy.` 

=> 4 new resources (2 x PVs and PVCs) and the local file used for reporting is created + destroyed

PR #860, opened 2024-10-15
feat(infra.ci) create a dedicated azure application and sp for updatecli read vm images

smerle33:
as per https://github.com/jenkins-infra/packer-images/pull/1443 from https://github.com/jenkins-infra/helpdesk/issues/4334
we need an azure credential to be able to run az commands in updatecli manifest on packer images.

we scoped this sp to a dedicated empty resource group to avoid a too extensive area of permissions. see https://learn.microsoft.com/en-us/azure/role-based-access-control/scope-overview for reference


dduportal (approved):
`Plan: 6 to add, 0 to change, 0 to destroy.`

LGTM

PR #866, opened 2024-10-22
feat(updates.jenkins.io) specify custom mount options for the SMB/CIFS ROX PVs

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4312#issuecomment-2429085702

This PR sets up custom mount options for the persistent volumes using SMB/CIFS fileshare, to limit side effects with the update of files "each 10 minutes" which may show files as present but empty.

- Default and recommended mount options are described in Azure documentation: https://learn.microsoft.com/en-us/azure/aks/azure-csi-files-storage-provision#mount-file-share-as-a-persistent-volume
- CIFS client options reference with explanations is there: https://manpages.debian.org/testing/cifs-utils/mount.cifs.8.en.html

NOTE: this change is not sufficient. Remount is required once deployed + Apache will need to be tuned.


smerle33 (approved):
LGTM

dduportal:
`Plan: 0 to add, 3 to change, 0 to destroy.`

PR #878, opened 2024-11-25
feat(updates.jenkins.io) add a new PV/PVC in RWX to allow rsync write…

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4402

The new RsyncD service aimed at replacing the `azcopy` service will have to write data, so we want a new `RWX` persistent volume.

This new PV/PVC will also replace the 2 existing (long term) with sub-directories (so we will provision only 1 time a premium 100 Gb instead of 2 today).


dduportal:
`Plan: 3 to add, 0 to change, 0 to destroy.` => 1 new file share, 1 new PV and 1 new PVC

PR #890, opened 2024-11-27
feat(publick8s): add a staging fileshare for geoip-data

smerle33:
provide a staging storage share to test the cronjob for geoip update without impacting the production storage share (fileshare)

### DIFF
--- a/publick8s.tf
+++ b/publick8s.tf
@@ -370,7 +370,7 @@ module "cronjob_geoip_data_fileshare_serviceprincipal_writer" {
 
 resource "azurerm_storage_share" "geoip_data_staging" {
   name                 = "geoip-data-staging"
-  storage_account_id = azurerm_storage_account.publick8s.id
+  storage_account_name = azurerm_storage_account.publick8s.name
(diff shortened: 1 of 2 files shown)


publick8s.tf
```diff
@@ -367,6 +367,48 @@ module "cronjob_geoip_data_fileshare_serviceprincipal_writer" {
...
+  }
+
+  data = {
+    azurestorageaccountname = azurerm_storage_account.publick8s.name
+    azurestorageaccountkey  = azurerm_storage_account.publick8s.primary_access_key
+  }
+
+  type = "Opaque"
+}
+
```
dduportal, line 398:
```suggestion
```

=> uneeded as the secret is scoped to the storage account (e.g. it's the same for all file shares of a given storage account)

dduportal (changes requested):
Not needed Kube namespace and Kube secrets

publick8s.tf
smerle33: see commit 16de42e

dduportal (approved):
`Plan: 7 to add, 0 to change, 0 to destroy.`

LGTM

PR #892, opened 2024-11-29
fix(trusted.ci/ephemeral-agents) add NSG rules to allow access the PE

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4402#issuecomment-2506556409

The goal is to fix the error `ssh: connect to host updates.jenkins.io-data.trusted.ci.jenkins.io port 22: Connection timed out`


dduportal:

PR #905, opened 2024-12-18
fix(db,updates.jio) correct and document network accesses to allow terraform from admin machines through VPN (instead of SSH tunnels)

dduportal:
Following work and discoveries by @smerle33 in https://github.com/jenkins-infra/docker-openvpn/pull/375 and https://github.com/jenkins-infra/docker-openvpn/pull/376

This PR corrects and documents the required accesses to allow us running `terraform plan` (before sending PRs) and `terraform apply` (hotfix/emergencies) from our admin machines.

- DBs routing is now fixed and using the VPN: documenting that `/etc/hosts` need to be updated (instead of using SSH tunnels) or using private subnets (infra.ci agents)
- Adding back the Update Center storage account IP restriction to the admin public IPs to avoid errors

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -30,6 +30,10 @@ In order to use this repository to provision the Jenkins infrastructure on azure
 git submodule update --init --recursive
 ----
 
+* Routing to database (private endpoints and private DNSes):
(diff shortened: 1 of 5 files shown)


dduportal:

PR #906, opened 2024-12-18
feat(infra.ci.jenkins.io) allow WinRM from agents to anywhere

dduportal:
Related to https://github.com/jenkins-infra/packer-images/pull/1611#issuecomment-2552011686

This PR allows WinRM outbound to *any* internet IP, as we can't determine in advance the public IPs of packer images in EC2.


dduportal (commented):

PR #930, opened 2025-02-04
chore(outputs) export endpoints of publick8s and privatek8s

dduportal:
This PR adds the hostname of the 2 "legacy" clusters (`publick8s` and `privatek8s`) to the infra report.

The goal is to allow Puppet and OpenVPN to access these information to keep the routing (to control planes through VPN).


dduportal:
`Plan: 1 to add, 0 to change, 1 to destroy.` => LGTM (only the output)

PR #940, opened 2025-02-19
 feat(ldap) provision statically the LDAP backup Azure File volume 

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4544#issuecomment-2671068585

This PR sets up the LDAP backup Azure File volume to provide a statically provisioned PVC.
It uses the same technique as updates.jenkins.io's GeoIPdata volume which is the same kind (except LDAP backup are ReadWriteMany while GeoIpdata are ReadOnlyMany).


smerle33 (approved):
looks good, I checked the namespace references and secrets metadata links ... everything sounds good.

I am assuming that 100Go is enough for that drive (l210)

dduportal:
Merging despite check failures:

- Done in pair with @smerle33 . We are watching carefully the deployment.
- Code reviewed in pair
- Plan carefully review: `Plan: 2 to import, 4 to add, 1 to change, 1 to destroy.`
- The check failure is only due to the imports, during the test harness step (not on the main terraform step)

PR #972, opened 2025-03-30
chore(output/publick8s/privatek8s)  export required outbound IP and CIDRs

dduportal:
This change add data sources to retrieve the effective outbound IPs (v4 and v6 when present) so that we can export these IPs.
It also exports the Pod CIDRs for `publick8s` 

=> the goal is to allow LDAP LB source allow-list to be updated properly


dduportal:
`Plan: 1 to add, 0 to change, 0 to destroy.` (the json report) LGTM

PR #974, opened 2025-03-31
chore(outputs) export puppet.jenkins.io in and out IPv4s for easier tracking

dduportal:
The `puppet.jenkins.io` VM is in the DMZ, its public IP is used for both inbound requests (SSH, puppet) and outbound requests (HTTP, LDAP).

As such, we want to export both in the infra reports so we can track the IPs where they are needed (tracking security group rules to puppet instead of relying on a git submodule, allowing in LDAP, etc.)


dduportal:
`Plan: 1 to add, 0 to change, 1 to destroy.` LGTM

PR #979, opened 2025-04-01
fix(cijenkinsio-agents-1) create the missing admin SA

dduportal:
Now that the network routing is fixed, we can proceed with the SA re-creation, so that infra.ci may manage the cluster.

Ref. https://github.com/jenkins-infra/helpdesk/issues/4619#issuecomment-2769875877


dduportal:
`Plan: 3 to add, 1 to change, 0 to destroy.` (the changed resource is the cluster, unrelated to this PR, ref. https://github.com/jenkins-infra/helpdesk/issues/4619#issuecomment-2768847941)

PR #975, opened 2025-03-31
feat(ci.jio-agent-1) migrating back cluster to azure

smerle33:
WIP

as per https://github.com/jenkins-infra/helpdesk/issues/4618
depends on https://github.com/jenkins-infra/azure-net/pull/354

create cluster for ci.jenkins.io agents on azure

PR #951, opened 2025-02-27
cleanup(azure/ci.jio): remove unused ressources migrate to AWS

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4313#issuecomment-2675163487

cleaning old ressources

PR #980, opened 2025-04-02
hotfix(providers) uses the correct FQDN for the cluster cijenkinsio-agents-1

dduportal:
Fixup of #975 which did not reverted the changes on `providers.tf` from #951 .

That should allow the changes from #979 to pass


dduportal:
`Plan: 3 to add, 1 to change, 0 to destroy.` => same as #979  let's roll

PR #981, opened 2025-04-02
fix(AKS clusters) ensure we only outputs required management elements with the correct cluster hostnames

dduportal:
As shown by hotfixes we had to apply (#978 #979) related to the cluster FQDNs in kubeconfigs,
this PR is a tentative to clean up all of these elements in the 4 AKS clusters:

- Removing unneeded outputs
- Ensure we have a sensitive output for each "kubernetes-management" kubeconfig for the infra.ci SOPS credentials. This should be the only "kubeconfig" output: if a maintainer is unable to retrieve the kubeconfig of a given AKS cluster for their machine, then they should not be granted maintainer permissions
- Fix the "kubernetes-management" kubeconfig to use the private FQDN (same as provider). Had to create a local for this as we cannot read the value from provider inside the kubernetes_sa module
  - The local must be distinct from the current `aks_cluster` otherwise we have a cyclical error when terraform expands it.
- Factorize the "input" parameters for clusters to the local


smerle33 (approved):
LGTM

PR #982, opened 2025-04-02
chore(cijioagents1) set default values for upgrade

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4619#issuecomment-2768847941

set the default value as code


dduportal (approved):

PR #983, opened 2025-04-02
Delete `cijenkinsio-agents-1`

dduportal:
Cleanup, restarting the cluster creation

Fixup of #975 as it was originated from a wrong source file which led to many issues due to missing code


dduportal:
```
Plan: 1 to add, 0 to change, 1 to destroy.
```
=> the infra report.

Along with the 2 outputs removed

PR #985, opened 2025-04-03
feat(ci.jenkins.io/cijenkinsio-agents-1) create required storage and PV/PVCs for maven cacher

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4619

This PR creates the storage account with required network restrictions (hence adding the missing subnets), fileshare (premium SMB at 100 Gb), and the 3 couples of PV/PVC, along with the outputs

PR #990, opened 2025-04-03
hotfix(ci.jenkins.io) use correct provider for the `maven-cache` Azure File Share

dduportal:
Fixup of #985 #987 and #988

Ref. https://github.com/jenkins-infra/helpdesk/issues/4619

The failure to locate the azure file comes from a missing `provider` on the Azure File Share resource in the TF code.
The share has been created but cannot be read as it searches in a distinct subscription (Azure CDF instead of Azure Sponsored).


dduportal:
`Plan: 8 to add, 0 to change, 0 to destroy.`

PR #996, opened 2025-04-04
fix(ci.jenkins.io) correct SSH inbound rules to allow VPN routing through the public outbound interface

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4620#issuecomment-2778047717


dduportal:
`Plan: 0 to add, 1 to change, 0 to destroy.` as expected

PR #999, opened 2025-04-05
fix(ci.jenkins.io) move to system identity instead of distinct Azure AD SP

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4620

Requires https://github.com/jenkins-infra/shared-tools/commit/f27614a0581a0c96284b86390c486720eec6b29e which disables non system identity Azure AD resource (AD App, SP and SP password).


dduportal:
`Plan: 5 to add, 0 to change, 7 to destroy.`

PR #1000, opened 2025-04-07
feat(ci.jenkins.io) improve workload identity setup and outputs

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4620 and specifically https://github.com/jenkins-infra/helpdesk/issues/4628#issuecomment-2782788349

Requires the 2 following modules changes:

- Deprecates the boolean "use system identity" in favor of the presence/absence of the SP end date (absent on ci.jio, present on the other controllers): https://github.com/jenkins-infra/shared-tools/commit/925aa14b886b21f9db45b68ca47f621963b36f67
- Add more outputs: https://github.com/jenkins-infra/shared-tools/commit/4d27dbb98ff2af2ee07e7fef5ee89975a26a4060

### DIFF
--- a/ci.jenkins.io.tf
+++ b/ci.jenkins.io.tf
@@ -26,7 +26,6 @@ module "ci_jenkins_io_sponsorship" {
   is_public                    = true
   enable_public_ipv6           = true
   default_tags                 = local.default_tags
-  enable_vm_system_identity    = true
 
   controller_resourcegroup_name = "ci-jenkins-io"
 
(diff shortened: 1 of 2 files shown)


dduportal:
`Plan: 1 to add, 0 to change, 2 to destroy.`

- 1 removal is the former Azure AD (Entra) Application for ci.jenkins.io (which was needed for the SP and its password, but not needed anymore with System identity)
- 1 add +  1 removal is the report file with new outputs
- Note: the Azure AD apps for cert.ci and trusted.ci are automatically moved (internally to Terraform) due to the introduction of a `count` attribute in their module definition

PR #1002, opened 2025-04-08
fix(cijenkinsio-agents-1) use the corect `role` label for the BOM node pool

dduportal:
ref. https://github.com/jenkins-infra/helpdesk/issues/4620

Otherwise BOM pods do not trigger the autoscaler because of label mismatch with pod template node selectors


dduportal:
`Plan: 0 to add, 1 to change, 0 to destroy.` let's roll

PR #1001, opened 2025-04-07
feat(cert.ci): enable system identity / workload identity to avoid credentials rotations

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4629
setting the VM to user systemidentity

depends on https://github.com/jenkins-infra/shared-tools/commit/c044f171db64ab6391a9401c6837f80b1c06c5ab

PR #1010, opened 2025-04-22
feat(trusted.ci): switch to workload identity

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4630

and following https://github.com/jenkins-infra/azure/pull/1001/files

remove the service principal and switch to workload identity for trusted.


dduportal (approved):
`Plan: 9 to add, 1 to change, 12 to destroy.`

- 9 role assignments which are destroyed and re-created (as their subject changed)
- 3 AzureAD (app, password and SP) resources destroyed
- 1 resource changed (The VM which has its system identity enabled)

PR #1016, opened 2025-04-23
fix(privatek8s-sponsorship) correct hostname for AKS API to use internal DNS

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4250#issuecomment-2824642025

This change applies the same technique as for the provider. We have to use the DNS hostname (instead of the "privatelink" default hostname). Same as what we did for cijenkinsioagents1 and infraciagents1


smerle33 (approved):
👍

PR #1017, opened 2025-04-24
fix(privatek8s-sponsorship) add missing 'Network Contributor' cluster permissions on subnbets

dduportal:
(need to be Network Contributor on all subnets)

Ref. https://github.com/jenkins-infra/helpdesk/issues/4250#issuecomment-2824642025

Required for the private Nginx ingress's LB


dduportal:
`Plan: 4 to add, 0 to change, 1 to destroy.`, tested with success

PR #1031, opened 2025-05-02
feat(cert.ci.jenkins.io) use a module for LetsEncrypt DNS setup: stricter authorization and reusability

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4629#issuecomment-2846851326

As per https://go-acme.github.io/lego/dns/azuredns/index.html#azure-managed-identity-with-azure-workload, we can restrict permissions even further.

This is the goal of this PR, along with defining resources into a reusable module to avoid risks of copy and pasting when applying to trusted.ci.

Requires https://github.com/jenkins-infra/shared-tools/commit/4542811deb41dbb5ea0d146c01f0c588141232b5

### DIFF
--- a/cert.ci.jenkins.io.tf
+++ b/cert.ci.jenkins.io.tf
@@ -7,7 +7,7 @@ module "cert_ci_jenkins_io" {
     azuread     = azuread
   }
 
-  service_fqdn                 = azurerm_dns_zone.cert_ci_jenkins_io.name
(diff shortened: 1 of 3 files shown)


dduportal:
`Plan: 2 to add, 0 to change, 1 to destroy.`

- 2 existing resources moved (0 to change): the DNZ zone and the parent NS record
- 1 resource to add (the new restricted permission for the ACME TXT record)
- 1 resource to recreate (existing moved permission, recreated due to role change)

PR #1032, opened 2025-05-02
feat(trusted.ci.jenkins.io) set up Let's Encrypt with Azure DNS and System/Workload identity

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4630#issuecomment-2846886448

This PR sets up trusted.ci.jenkins.io VM to use credential-less Let's Encrypt DNS azure with restricted permissions in the distinct DNS zone.

- Imports the existing DNS zone and parent NS records (removed from azure-net in https://github.com/jenkins-infra/azure-net/pull/377)
- A bit of factorization with the FQDN


dduportal:
`Plan: 2 to import, 2 to add, 2 to change, 0 to destroy.`

- 2 imported resources as expected...
- ... which are both changed: their tag `repository` is set to this repository (instead of former azure-net)
- And 2 new permissions schemes added to the VM identity
🚀

PR #1041, opened 2025-05-20
cleanup(uplink) temporarily stop managing the DB

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4667#issuecomment-2894701228

The goal is to avoid accidentally removing/changing the schema and role during migrations.


dduportal:
`Plan: 0 to add, 0 to change, 0 to destroy.`, the 3 resources are marked as `no longer be managed by Terraform, but will not be destroyed`

PR #1074, opened 2025-06-20
feat(reports.jenkins.io) create a new storage account, new RG and also a file share for `builds.reports` with managed identity

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4696

### DIFF
--- a/reports.jenkins.io.tf
+++ b/reports.jenkins.io.tf
@@ -1,4 +1,6 @@
 ## This file contains the resources associated to the buckets used to store private and public reports
+
+############# Legacy resources to be removed once migrated to the new resources below
 resource "azurerm_resource_group" "prod_reports" {
   name     = "prod-reports"
   location = var.location
@@ -7,7 +9,6 @@ resource "azurerm_resource_group" "prod_reports" {
     scope = "terraform-managed"
   }
 }
-
 resource "azurerm_storage_account" "prodjenkinsreports" {
   name                       = "prodjenkinsreports"
   resource_group_name        = azurerm_resource_group.prod_reports.name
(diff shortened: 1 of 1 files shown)


dduportal:
`Plan: 8 to add, 0 to change, 0 to destroy.` let's roll!

dduportal:
Fail to apply 2 network ACLs on the new storage account with the error `SubnetsHaveNoServiceEndpointsConfigured` for the subnet `cert-ci-jenkins-io-sponsorship-vnet-ephemeral-agents`.

=> we have to enable the `Microsoft.Storage` endpoint on the cert.ci agents subnets. See https://github.com/jenkins-infra/azure-net/pull/394

reports.jenkins.io.tf
```diff
@@ -22,3 +23,91 @@ resource "azurerm_storage_account" "prodjenkinsreports" {
...
+  provider = kubernetes.publick8s
+
+  metadata {
+    name = azurerm_storage_share.builds_reports_jenkins_io.name
+    labels = {
+      name = azurerm_storage_share.builds_reports_jenkins_io.name
+    }
+  }
+}
+resource "kubernetes_service_account" "builds_reports_jenkins_io" {
```
dduportal, line 100: I forgot to set a provider in this resource, which caused the error

```
│ Error: serviceaccounts is forbidden: User "system:serviceaccount:jenkins-infra-agents:default" cannot create resource "serviceaccounts" in API group "" in the namespace "builds-reports-jenkins-io"
```

("system:serviceaccount:jenkins-infra-agents:default" is the default service account in `infracijenkinsioagentsX` clusters)

PR #1079, opened 2025-06-22
fix(builds.reports.jenkins.io) use distinct set of resource than reports.jenkins.io to allow distinct identity permissions

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4696#issuecomment-2994110704

Note: the workload identity is only applied to infra.ci.jio's VM and container agents for now.

Tested this PR manually and validated it with 3 actions:

- Created a pod in `publick8s` which mounts the PVC (readonly) with success ✅ 
- With infra.ci.jenkins.io and a manual change to 
- Replayed a build in infra.ci.jio with the `linux-amd64` label (Linux VM agent) and ran the command `echo $(date) > test.txt; azcopy copy test.txt "https://buildsreportsjenkinsio.file.core.windows.net/builds-reports-jenkins-io/"`with success ✅ 
- Created a pod in `publick8s` which mounts the PVC (readonly) with success ✅

PR #1081, opened 2025-06-22
feat(cert.ci.jio/trusted.ci.jio) allow agents to write to builds.reports.jenkins.io

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4696

Apply the same VM agents identity technique as #1079 but for cert.ci and trusted.ci.


dduportal:
`Plan: 6 to add, 1 to change, 0 to destroy.` 

-  twice 3 new resource
- The trusted.ci permanent agent is changed to have identity

PR #1082, opened 2025-06-23
feat(infracijioagents2): set minimum autoscaler to 1 for arm64 nodepool

smerle33:
autoscaler seem to be a bit off with arm64 and nodepool starting at 0 ... as we use arm64 pods a lot, lets set it to 1 as min and check that the autoscaler works from here.


smerle33:
```
      ~ min_count                     = 0 -> 1

Plan: 0 to add, 1 to change, 0 to destroy.
```

PR #1087, opened 2025-06-25
fix(infra.ci/infracijenkins-agents-2) manage service account for infra.ci.jio container agent

dduportal:
To ensure federated identity works as expected, we need to explicitly specify a service account into our kubernetes pod agent templates: It's done in https://github.com/jenkins-infra/kubernetes-management/pull/6727.

However, since the service account requires an annotation with the User Assigned Identity and the federated credential needs the service account name, then it sounds way more coherent to manage the aforementioned SVC account in Terraform here and export its name for kubernetes-management. 

Ref. https://github.com/jenkins-infra/helpdesk/issues/4696#issuecomment-3000046567

Note: this PR also updates the reports content and may break some downstream `updatecli` manifests. I'll check them (mostly puppet and kubernetes AFAICT)


dduportal:
`Plan: 2 to add, 1 to change, 1 to destroy.`

PR #1098, opened 2025-07-08
cleanup(ci.jenkins.io) remove cijenkinsioagents-1 Kubernetes resources

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4688

This is the first cleaning up steps. Need to be a 2 times process so that we can safely remove the Kubernetes provider on the 2nd step.


dduportal:
`Plan: 1 to add, 0 to change, 15 to destroy.`

PR #1108, opened 2025-07-22
feat(privatek8s) migrate from sponsored to cdf account

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/4690

handover to @dduportal as seen in pair


dduportal (approved):
`Plan: 16 to add, 0 to change, 0 to destroy.` => let's roll with this first shot

PR #1111, opened 2025-07-23
cleanup(trusted.ci.jio) remove resources from the Azure Sponsored subscription

dduportal:
…


dduportal:
`Plan: 0 to add, 6 to change, 31 to destroy.` Let's roll!

PR #1125, opened 2025-07-24
 fix(infra.ci) allow controller UAID to read the packer images RG

dduportal:


dduportal:
`No changes. Your infrastructure matches the configuration.` => already applied and verified manually. Let's roll

PR #1162, opened 2025-08-14
hotfix(data-storage.jenkins.io) allow release.ci agents to access the new data storage

dduportal:
…


dduportal:
Self merging as already manually to verify it works as expected

PR #1171, opened 2025-08-19
feat(dockerhub-mirror) enable access to the ACR from trusted.ci.jenkins.io through private endpoint

dduportal:
…


dduportal:
Failed to apply: there is already a private DNS zone linked in the trusted.ci virtual network (used for mirrorbits endpoints).

=> as such we need to reuse this private DNS zone.

See https://github.com/jenkins-infra/azure/pull/1172

PR #1172, opened 2025-08-19
fix(dockerhub-mirror) use existing private DNS zone for trusted.ci 

dduportal:
Fixup of https://github.com/jenkins-infra/azure/pull/1171#issuecomment-3199486226

Note: terratest are annoying here as they show the weakness of terraform plan on an empty platform when a lot of "dynamic" loops are used. As such, we do not "add" some resources when running in terratest to avoid errors.

This PR also adds NSG rules for trusted <-> ACR (missing in #1171)


dduportal:
`Plan: 2 to add, 2 to change, 1 to destroy.`

PR #1181, opened 2025-09-10
chore(publick8s) rename resources with the prefix 'old_' to limit conflicts with the new cluster

dduportal:


dduportal:
`Plan: 0 to add, 0 to change, 0 to destroy.` Let's roll

PR #1187, opened 2025-09-16
feat(publick8s) create a new cluster with modern (2025) setup

dduportal:

This new cluster has a set of new modern features and other useful for the current jenkins-infra context:

- Default outbound method set to the subnet's NAT gateway (user defined NAT gateway) - https://github.com/jenkins-infra/helpdesk/issues/4617 
- CNI network used, same as `privatek8s` - https://github.com/jenkins-infra/helpdesk/issues/4617
- IPv6 dual stack enabled (same as before)
- Only 1 node pool which is the default one:
  - Using Azure Linux (instead of Ubuntu like we use to)
  - Mixing AKS Workload with ours (eg. no secondary node pool) to decrease the costs: we should have around 3 to 4 nodes (instead of 3 + 3 in the old `publick8s`). We don't have the constraint of separating these workloads in this cluster


dduportal:
`Plan: 12 to add, 0 to change, 0 to destroy.` Let's roll!

PR #1193, opened 2025-09-18
chore(oldpublick8s) rename and move Kubernetes NS/Secrets/PVs/PVCs to avoid conflicts with the new publick8s

dduportal:
…


dduportal:

PR #1194, opened 2025-09-19
feat(publick8s) create kubernetes resources for services with a new factorized syntax

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4617#issuecomment-3295780800

This PR creates most of the `publick8s` resources using iterator on a `local` data structure.

We can improve this in the future with modules (to reuse the logic across clusters) but I'm not sure if it is worth as we would want to get rid of `privatek8s` in favor of VMs (cheaper)


dduportal:
`Plan: 31 to add, 0 to change, 0 to destroy.` Let's roll! (maps to my local tests)

PR #1196, opened 2025-09-20
feat(publick8s) recreate new cluster with its own subnet

dduportal:
Requires https://github.com/jenkins-infra/azure-net/pull/436 (which creates the new subnet) and https://github.com/jenkins-infra/azure-net/pull/437 (setting up the subnet for modern setup as "private" without any outbound method)


dduportal:
`Plan: 6 to add, 0 to change, 0 to destroy.`

PR #1199, opened 2025-09-22
feat(publick8s) add custom PV/PVCs for all web services except www

dduportal:

- Add PV/PVCs (and NS + secrets) in CSI to existing web services shared accounts
- Create a new file share for reports.jenkins.io (start empty)

Note: all these PV/PVCs are expected to move (soon) to the centralized NFS: https://github.com/jenkins-infra/helpdesk/issues/4767. For now, let's migrate web services to the new cluster with explicit PV/PVCs (instead of pod direct CSI mounts in the old `publick8s-endless-ghoul` cluster)

### DIFF
--- a/locals.tf
+++ b/locals.tf
@@ -93,10 +93,7 @@ locals {
       azurefile_volumes = {
         "get-jenkins-io"     = {},
         "updates-jenkins-io" = {},
-        "javadoc-jenkins-io" = {},
         "www-jenkins-io"     = {},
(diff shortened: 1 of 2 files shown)


dduportal:
`Plan: 20 to add, 0 to change, 2 to destroy.`

- 18 new resources
- PV and PVC for javadoc are re-created because moved from NFS to existing storage share in CIFS

PR #1207, opened 2025-09-25
feat(publick8s-endless-ghoul) create custom DNS records for migration

dduportal:
and set up TTL of all records to 60s while we are there

Related to https://github.com/jenkins-infra/helpdesk/issues/4617#issuecomment-3332342227


dduportal:
`Plan: 2 to add, 3 to change, 0 to destroy.`

PR #1212, opened 2025-09-25
fix(publick8s,reports) use both new and old PVC with valid setup

dduportal:


locals.tf
```diff
@@ -220,8 +220,29 @@ locals {
           secret_namespace    = "plugins-jenkins-io",
           storage_account_key = azurerm_storage_account.plugins_jenkins_io.primary_access_key,
         },
+        "reports-jenkins-io-old" = {
```
dduportal, line 223:
```suggestion
        # TODO: delete when migrated to the new storage file share
        "reports-jenkins-io-old" = {
```

dduportal:
`Plan: 1 to import, 3 to add, 0 to change, 0 to destroy.`

PR #1225, opened 2025-10-06
feat(reports/infraci/trustedci) add SP to allow SAS token to write into the new reports.jenkins.io file share

dduportal:
…


dduportal:
`Plan: 12 to add, 0 to change, 0 to destroy.`

PR #1227, opened 2025-10-06
fix(reports) allow access from trusted.ci agents

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4821#issuecomment-3372038800

We forgot to allow trusted.ci agents for reports storage


dduportal:
`Plan: 1 to add, 1 to change, 1 to destroy.`

PR #1385, opened 2026-03-31
chore: cleanup unused locals, shared tools bump and formatting

dduportal:
Ref. https://github.com/jenkins-infra/azure/pull/1383#pullrequestreview-4036378454

>  looks like this local is not used anywhere.

This PR is a minor change which remove unused local values along with their updateCLI tracking to avoid unneeded PRs like #1383.

Note: I also bumped the `.shared-tool` submodule (to ensure no `trivy` is used by our local machines) and applied a bit of formatting since these are non functional changes. I did separated commits to ease the review though.


dduportal:
`No changes. Your infrastructure matches the configuration.` => Self merging as it's a non functional change.

PR #1390, opened 2026-04-14
chore: create a dummy agent in the same resource groupe as the current permanent agent

lemeurherve:
This change creates a dummy agent in the same `permanent-agents-trusted-ci-jenkins-io` RG as the current permanent agent in the CDF subscription to test its move to the Sponsored subscription via Azure Portal.

> Plan: 7 to add, 0 to change, 0 to destroy.

Refs:
- https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/move-resource-group-and-subscription

### DIFF
--- a/dummy.trusted.ci.jenkins.io.tf
+++ b/dummy.trusted.ci.jenkins.io.tf
@@ -1,13 +1,8 @@
 ####################################################################################
 ## Resources for the permanent agent VM
 ####################################################################################
-resource "azurerm_resource_group" "permanent_agents_trusted_ci_jenkins_io" {
-  name     = "permanent-agents-trusted-ci-jenkins-io"
-  location = var.location
-  tags     = local.default_tags
-}
-resource "azurerm_network_interface" "agent_trusted_ci_jenkins_io" {
-  name                = "agent-trusted-ci-jenkins-io"
+resource "azurerm_network_interface" "dummy_trusted_ci_jenkins_io" {
+  name                = "dummy-trusted-ci-jenkins-io"
   location            = azurerm_resource_group.permanent_agents_trusted_ci_jenkins_io.location
   resource_group_name = azurerm_resource_group.permanent_agents_trusted_ci_jenkins_io.name
   tags                = local.default_tags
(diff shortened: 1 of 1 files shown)


dummy.trusted.ci.jenkins.io.tf
```diff
@@ -0,0 +1,117 @@
...
+  caching            = "None" # Caching not supported with "PremiumV2_LRS"
+}
+
+####################################################################################
+## Network Security Group and rules
+####################################################################################
+resource "azurerm_subnet_network_security_group_association" "trusted_ci_dummy_agent" {
+  subnet_id                 = data.azurerm_subnet.trusted_ci_jenkins_io_permanent_agents.id
+  network_security_group_id = module.trusted_ci_jenkins_io.controller_nsg_id
+}
```
dduportal:
```suggestion
```

Duplicates https://github.com/jenkins-infra/azure/blob/791e25baf1baaa089bace96ed91f23a8342facf1/agent.trusted.ci.jenkins.io.tf#L92-L95

dduportal (dismissed):
Giving approval but need my suggestion to be applied (duplicated subnet.NSG association) otherwise it most probably will fail during apply

dduportal (approved):

dummy.trusted.ci.jenkins.io.tf
```diff
@@ -0,0 +1,113 @@
...
+  caching            = "None" # Caching not supported with "PremiumV2_LRS"
+}
+
+####################################################################################
+## Network Security Group and rules
+####################################################################################
+
+resource "azurerm_network_security_rule" "allow_inbound_ssh_from_controller_to_dummy_agent" {
+  name                   = "allow-inbound-ssh-from-controller-to-dummy-agent"
+  priority               = 3600
```
lemeurherve, line 90:
```suggestion
  priority               = 3601
```

Distinct priority applied as hotfix in https://github.com/jenkins-infra/azure/commit/1662a4bfac999958cff396a49da291b1cc547459

PR #1391, opened 2026-04-15
chore(dummy): remove public_key metadata

lemeurherve:
Small test to see if changing that public key metadata triggers a VM reprovision.


lemeurherve:
The plan confirms it would trigger a VM reprovision, closing.

> Plan: 3 to add, 2 to change, 2 to destroy.

(cc @dduportal FYI)

PR #1393, opened 2026-04-16
chore(dummy): remove resources from state prior to their Portal migration

lemeurherve:
This change removes `dummy` resources from terraform state prior to their migration via Azure Portal from the current permanent agent RG in CDF subscription to the new RG created in the Sponsored subscription cf https://github.com/jenkins-infra/azure/pull/1392.

Ref:
- https://developer.hashicorp.com/terraform/language/state/remove
- https://github.com/jenkins-infra/helpdesk/issues/5067

PR #1394, opened 2026-04-16
chore(dummy): reimport resources into state

lemeurherve:
This change re`import`s resources removed from terraform state in #1393, except the data disk which I moved to the sponsored subscription via Azure Portal, a new one will be recreated.
Similarily, the disk attachment hasn't been `import`ed as the corresponding disk will be a new one.

Ref:


lemeurherve:
> Plan: 4 to import, 2 to add, 0 to change, 0 to destroy.

LGTM, merging.

lemeurherve:
All good:
> Apply complete! Resources: 4 imported, 2 added, 0 changed, 0 destroyed.

PR #1396, opened 2026-04-16
chore(modules) migrate shared module `azure-jenkinsinfra-azurevm-agents` into local repository

dduportal:
The goal is to avoid bad surprises when changing the shared code module (which is used by default on the main branch). Instead, we open a PR with required changes and we will viw, review and approve/refuse the real terraform plan (instead of running a build on main and expecting surprises when the change has been pushed to the shared-tools repository).

This pattern has been successfully tested on azure-net with the `azure-vnet` module and is allowed because the module `azure-jenkinsinfra-azurevm-agents` is only used here.

See https://github.com/jenkins-infra/helpdesk/issues/5070#issuecomment-4260751840 for details.


dduportal:

lemeurherve (approved):
We probably want to also import the GitHub IP adresses updatecli manifest: https://github.com/jenkins-infra/shared-tools/blob/main/updatecli/updatecli.d/git-inbound-ips.yml

(Worth renaming it with `github` instead of `git` at the same time)

dduportal:
> We probably want to also import the GitHub IP adresses updatecli manifest: https://github.com/jenkins-infra/shared-tools/blob/main/updatecli/updatecli.d/git-inbound-ips.yml
> 
> (Worth renaming it with `github` instead of `git` at the same time)



> We probably want to also import the GitHub IP adresses updatecli manifest: https://github.com/jenkins-infra/shared-tools/blob/main/updatecli/updatecli.d/git-inbound-ips.yml
> 
> (Worth renaming it with `github` instead of `git` at the same time)

Yep, totally. Will be a subsequent PR here :)

PR #1397, opened 2026-04-16
feat!(modules/azure-jenkinsinfra-azurevm-agents) allow using common NSG from Virtual Network Resource Group with same name

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5070#issuecomment-4260751840

This PR introduces a new input variable on the module `azure-jenkinsinfra-azurevm-agents`: `use_vnet_common_nsg` which defaults to false (opt-in).

This input tells the module to use an existing NSG, in the resource group of the provided virtual network, with the same name as the provided virtual network. It is a convention to ensure we stay coherent.
If we need a custom name, we'll change this new input from a boolean to a string, but does not seem useful for now.

It will allow migrating our NSGs with as less impact as possible.

BREAKING: the input parameter `controller_rg_name` is renamed to `nsg_rg_name` as it's its only usage.


dduportal:

lemeurherve (approved):
🚀

PR #1402, opened 2026-04-17
feat!(trusted.ci.jenkins.io/sponsored-subscription) migrate ACR Private Endpoint and its Network Security rules to the commons subnet and Resource Group

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5070#issuecomment-4268303462

This PR re-create the following resources:
- The Private Endpoint used to reach the Azure Container Registry from trusted.ci.jenkins.io agents in the sponsored virtual network subnets. The new NIC for this endpoint will live in the "commons" subnet instead so we can dispose of the other subnets and avoid interferences during migrations/updates/cleanups
- The 2 NSG rules to allow request from/to the private endpoints:
  - Naming changed to shows it's "sponsored"
  - Priority changed to avoid "deploy time" error requiring 2 consecutive "apply"s
  - NSG and its RG are defined from the data source to remove dependency to the ephemeral (sponsored) module instanciation

⚠️ When this PR will be applied, the trusted.ci.jenkins.io ephemeral agents will loose connectivity to the private ACR. Not a blocker as their Docker Engine will fallback to the DockerHub (a bit slower and a bit of bandwidth).

Note: review should only be done on the second commit (https://github.com/jenkins-infra/azure/pull/1402/changes/92ed414b0567982b42392258db728b01c9fbeaa3) as the first one (https://github.com/jenkins-infra/azure/pull/1402/changes/f848f104144854f93ebe6e265a10a7e0ed575e5b) is only a formatting one (moving definitions in the file)

### DIFF
--- a/trusted.ci.jenkins.io.tf
+++ b/trusted.ci.jenkins.io.tf
@@ -66,6 +66,66 @@ resource "azurerm_resource_group" "trusted_ci_jenkins_io_sponsored_commons" {
   location = var.location
   tags     = local.default_tags
 }
+# Managed in jenkins-infra/azure-net with vnet and subnets
+data "azurerm_network_security_group" "trusted_ci_jenkins_io_sponsored_vnet" {
+  provider = azurerm.jenkins-sponsored
+
+  name                = "trusted-ci-jenkins-io-sponsored-vnet"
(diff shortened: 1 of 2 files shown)


dduportal:
`Plan: 3 to add, 0 to change, 3 to destroy.` => looks good and expected. Requesting review before merging

lemeurherve (approved):

PR #1403, opened 2026-04-17
chore: create dummy2

lemeurherve:
Ref:
- https://github.com/jenkins-infra/helpdesk/issues/5067#issuecomment-4261189823

PR #1405, opened 2026-04-17
chore: upgrade dummy data disk to ZRS

lemeurherve:
Upgrading to allow attaching this disk to the dummy2 VM which isn't in the same zone.

Fixup of:
- #1403 
- #1404


lemeurherve:
> Error: expected storage_account_type to be one of ["Standard_LRS" "StandardSSD_ZRS" "Premium_LRS" "PremiumV2_LRS" "Premium_ZRS" "StandardSSD_LRS" "UltraSSD_LRS"], got PremiumV2_ZRS

PR #1414, opened 2026-04-20
feat: add agent-2.trusted.ci.jenkins.io

lemeurherve:
Changes to review: https://github.com/jenkins-infra/azure/pull/1414/changes/e8c802611508e1b0a7f2cd4ac0b4216b3e6a2164..HEAD

Ref:


agent-2.trusted.ci.jenkins.io.tf
```diff
@@ -25,31 +20,32 @@ resource "azurerm_network_interface" "agent_trusted_ci_jenkins_io" {
...
-  resource_group_name             = azurerm_resource_group.permanent_agents_trusted_ci_jenkins_io.name
-  location                        = azurerm_resource_group.permanent_agents_trusted_ci_jenkins_io.location
+resource "azurerm_linux_virtual_machine" "agent_2_trusted_ci_jenkins_io_jenkins_sponsored" {
+  provider                        = azurerm.jenkins-sponsored
+  name                            = "agent-2.trusted.ci.jenkins.io"
+  resource_group_name             = azurerm_resource_group.trusted_ci_jenkins_io_permanent_agents_jenkins_sponsored.name
+  location                        = azurerm_resource_group.trusted_ci_jenkins_io_permanent_agents_jenkins_sponsored.location
   tags                            = local.default_tags
-  size                            = "Standard_B2s"
+  size                            = "Standard_D2s_v3" # 2 vCPU 8Go RAM
```
dduportal:
```suggestion
  size                            = "Standard_D2as_v7"
```

- The `DSv3` family is old. Not deprecated (yet?) but old serie: https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/general-purpose/dsv3-series?tabs=sizebasic
- Let's use a modern instance in v7 family.I 'm keeping your sizing choice (2 vCPU / 8 Gb). Note we might want to increase CPU instead of memory, but worth checking later in datadog (still can be changed later)
- Chose `Dasv7` instead of the usual `Dadsv7` we use on the new subscription: no need for local NVMe (data disk is Premium v2 SSD).

agent-2.trusted.ci.jenkins.io.tf
```diff
@@ -0,0 +1,129 @@
...
+}
+
+####################################################################################
+## Network Security Group and rules
+####################################################################################
+resource "azurerm_subnet_network_security_group_association" "trusted_ci_permanent_agent_2_jenkins_sponsored" {
+  provider                  = azurerm.jenkins-sponsored
+  subnet_id                 = data.azurerm_subnet.trusted_ci_jenkins_io_sponsored_permanent_agents.id
+  network_security_group_id = module.trusted_ci_jenkins_io.controller_nsg_id
+}
```
dduportal:
```suggestion
```

To be removed: we have to use the "common" NSG defined in azure-net: https://github.com/jenkins-infra/azure-net/pull/519.

The association subnet <-> NSG is managed in azure-net and unneeded here.

agent-2.trusted.ci.jenkins.io.tf
```diff
@@ -0,0 +1,129 @@
...
+
+resource "azurerm_network_security_rule" "allow_inbound_ssh_from_controller_to_permanent_agent_2_jenkins_sponsored" {
```
dduportal, line 98: Not blocking (I'll propose a naming convention later, you have to move forward on this VM setup): the wording `_controller_` in the old resources where your code comes from was badly chosen (I guess it was by me 😮‍💨 ).

We don't know, when looking at the list of Terraform resources in its state, if it is cert.ci, trusted.ci or other controller.

Also, the `permanent` wording seems a poor choice (again, most probably me) as it does not help to find the Terraform `tf` where it is defined. Since we have limit in number of characters for the name in Azure API, I would suggest to remove (again, I'll do it on a subsequent PR).

agent-2.trusted.ci.jenkins.io.tf
```diff
@@ -0,0 +1,129 @@
...
+  access                 = "Allow"
+  protocol               = "Tcp"
+  source_port_range      = "*"
+  destination_port_range = "22"
+  source_address_prefix  = module.trusted_ci_jenkins_io.controller_private_ipv4
+  destination_address_prefixes = [
+    azurerm_linux_virtual_machine.agent_2_trusted_ci_jenkins_io_jenkins_sponsored.private_ip_address,
+  ]
+  resource_group_name         = module.trusted_ci_jenkins_io.controller_resourcegroup_name
+  network_security_group_name = module.trusted_ci_jenkins_io.controller_nsg_name
```
dduportal:
```suggestion
  resource_group_name         = data.azurerm_network_security_group.trusted_ci_jenkins_io_sponsored_vnet.resource_group_name
  network_security_group_name = data.azurerm_network_security_group.trusted_ci_jenkins_io_sponsored_vnet.name
```

We have to use the common NSG now

agent-2.trusted.ci.jenkins.io.tf
lemeurherve: Missed the "previous" and "retirement" section names on the left of Azure doc 🤦 😅 

> image

dduportal (changes requested):
Good work.

- 1 main blocker: we have to use the "common" NSG now (see suggestions)
- 1 non blocker but important: VM size consideration, to start with a modern family
- 1 nit I might probably discuss with you later and I'll send a renaming PR (so you can move forward)

agent-2.trusted.ci.jenkins.io.tf
lemeurherve: I'll rework my PR with proper names, easier than coming back to it.

dduportal: I would prefer getting this PR deployed and you moving forward on the valuable tasks. Naming is a high nitpicking and does not deserve time from us until we've set up the first Puppet run. It's a distraction at best. I noted it as food for though, not because it creates a problem

lemeurherve: OK, going forward,

> Plan: 6 to add, 0 to change, 0 to destroy.

Merging this PR.

lemeurherve:
Failing with:

> Network Interface Name: "agent-2-trusted-ci-jenkins-io"): performing CreateOrUpdate: unexpected status 400 (400 Bad Request) with error: InvalidResourceReference: Resource /subscriptions/****/resourceGroups/trusted-ci-jenkins-io/providers/Microsoft.Network/virtualNetworks/trusted-ci-jenkins-io-vnet/subnets/trusted-ci-jenkins-io-vnet-permanent-agents referenced by resource /subscriptions/1e7d5219-acbc-4495-8629-bdbb22e9b3ed/resourceGroups/trusted-ci-jenkins-io-sponsored-permanent-agents/providers/Microsoft.Network/networkInterfaces/agent-2-trusted-ci-jenkins-io was not found. Please make sure that the referenced resource exists, and that both resources are in the same region.

Looking.

PR #1426, opened 2026-04-29
chore((cert|trusted).ci.jenkins.io) migrate `azure-jenkinsinfra-controller` module out of shared tools

dduportal:

The goal is to avoid bad surprises when we change something in the module: since we only use the module in this repository, a plain old PR will show a real life Terraform plan.

Alternative would be to publish module with versions: PRs with version bumps would have the same result (but more steps so only worth when used on many repositories).


dduportal:

PR #1429, opened 2026-04-29
feat(modules/azure-jenkinsinfra-controller) allow specifying the existing "common" NSG for all subnets of the controller's vnet

dduportal:

- Expect zero resources changed in the current state
- Tested with a new initialized module `trusted_ci_jenkins_io_sponsored` and the new variable set to true: result looks good


dduportal:

PR #1430, opened 2026-04-30
feat(trusted.ci.jenkins.io) instantiate a new controller VM in the sponsored subscription

dduportal:

This PR instantiates a new VM for trusted.ci.jenkins.io controller with the following changes from the current VM:

- Disk with the "default" convention name
- RG set to custom value following the agents RG names in the sponsored subscription
- Switching to the Standard Dasv6 family (TPM, faster IOs, modern bootloader, quotas available, more memory due to CPU/memory ratio, no more noisy neighbourns)
- New SSH key for user `jenkins-infra-team`
- Use the "common" NSG from the virtual net
- Custom hostname: `controller-sponsored.trusted.ci.jenkins.io` to avoid collisions

### DIFF
--- a/trusted.ci.jenkins.io.tf
+++ b/trusted.ci.jenkins.io.tf
@@ -55,6 +55,51 @@ module "trusted_ci_jenkins_io" {
 }
 
 ####################################################################################
+## Resources for the Controller VM in the sponsored subscription
+####################################################################################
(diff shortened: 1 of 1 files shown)


dduportal:
`Plan: 16 to add, 0 to change, 0 to destroy.` Let's roll

PR #1432, opened 2026-04-30
fix(modules/azure-jenkinsinfra-controller) ensure NSG rules are created in the resolved NSG RG (not always the controller)

dduportal:
Fixup of #1429 to solve https://github.com/jenkins-infra/azure/commit/ab0463168cce7d1aba9b1a9d2cb68b5599d9c0cf#r183931644

Related to https://github.com/jenkins-infra/helpdesk/issues/5084#issuecomment-4343922083

PR #1435, opened 2026-05-01
fix(modules/azure-jenkinsinfra-controller) create NSG rules with a customizable offset in the external NSG and fix nonsense rules only in this case

dduportal:
Fixup of #1432 and #1429

Many module changes due to the paradigm change from "one NSG per subnet" to "one NSG for the whole subnet"
Most of the fixes are only for the new paradigm (e.g. only [controller-sponsored.trusted.ci.jenkins.io](https://github.com/jenkins-infra/helpdesk/issues/5084) for now, cert.ci will follow) to avoid unwanted changes on current controllers.

- Introduces 2 variable offsets with a default value at 4000 to avoid priority collisions (and allow tuning)
- No need to create the default "deny_all" rules when the NSG is external: we expect these rules to be created in azure-net (not the case yet but soon)
- We don't want "open" inbound protocol for agents
- Naming of resources which requires discrimination to avoid naming collisions

### DIFF
--- a/modules/azure-jenkinsinfra-controller/main.tf
+++ b/modules/azure-jenkinsinfra-controller/main.tf
@@ -182,18 +182,15 @@ resource "azurerm_virtual_machine_data_disk_attachment" "controller_data" {
 ####################################################################################
 ## Network Security Group and rules
 ####################################################################################
-### Controller
 resource "azurerm_network_security_group" "controller" {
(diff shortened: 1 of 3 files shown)


dduportal:
`Plan: 6 to add, 0 to change, 5 to destroy.` => only changes on the sponsored trusted.ci controller, other resources are moved internally

PR #1436, opened 2026-05-02
feat(modules/azure-jenkinsinfra-controller) allow creating DNS record in the public zone even for private controllers

dduportal:
While working on https://github.com/jenkins-infra/helpdesk/issues/5084#issuecomment-4343922083, it appeared we are missing DNS `A` record resolving to our private controller IPv4. 

That the module was designed for ci.jenkins.io which is public. Back in that time, we were setting up our `/etc/hosts`. But since then, we've defined Letsencrypt for private controller using their own DNS zones: we can safely follow the AKS pattern of public DNS records resolving to private IUPv4 routed through VPN.

This change allows (if applied by setting `var.dns_zone_name` and `var.dns_resourcegroup_name`) for each controller to get a public DNS `A` record resolving to its private IPv4 (VPN routed).

No changes expected for this first PR: will create record on a subsequent PR.


dduportal:

PR #1446, opened 2026-05-04
fix(trusted.ci.jenkins.io) add missing NSG rule to allow HTTP inbound for admins through VPN

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5084#issuecomment-4344122327

Caused by https://github.com/jenkins-infra/azure-net/pull/527

Moving the "default" NSG rules to common NSG and enforcing them had a few side effect such as blocking HTTP access to the controller.

Worry not: this PR should fix this.


dduportal:

PR #1447, opened 2026-05-04
fix(agent-2.trusted.ci.jenkins.io) add missing NSG rules from the ephemeral_agents module

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5084#issuecomment-4344122327

Caused by https://github.com/jenkins-infra/azure-net/pull/527

Moving the "default" NSG rules to common NSG and enforcing them had a few side effect such as blocking outbound HTTP/git/HKP


dduportal:
`Plan: 4 to add, 0 to change, 0 to destroy.`

PR #1458, opened 2026-05-21
feat: create privatek8s azurerm (e.g. non kubernetes) resources

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/5091#issuecomment-4381409952

Notes:

- Added the kubernetes provider, so reverting PRs will be easier when decomissioning
- Already set up the required permissions for infra.ci and release.ci controllers
- Trying to create the public IP and its lock in the sponsored subscription (parent existing RG is in the CDF subscription but I assume the resources should be in the same subscription as the LB which will use it). Might need to revisit eventually
- Added a few comments around subscriptions and naming constraints in Azure API

PR #1464, opened 2026-05-21
fix(privatek8s-sponsored) Windows 2022 node pool are restricted to Gen1 VMs

dduportal:
Amends #1558 to fix the error described in https://github.com/jenkins-infra/azure/pull/1458/changes#r3282577059

Related to https://github.com/jenkins-infra/helpdesk/issues/5091#issuecomment-4381409952

This PR sets up the Windows 2022 Node pool of the new cluster to an older VM family (same as current cluster: `_v5`) because Generation 2 requires Windows 2025 in the context of AKS.

Despite MS documentation sayin g otherwise: https://learn.microsoft.com/en-us/azure/aks/generation-2-vms?tabs=windows-node-pool#create-a-node-pool-with-a-gen-2-vm,
the Terraform azurerm provider does not allow the custom "header" technique: https://github.com/hashicorp/terraform-provider-azurerm/issues/31526.

And of course Windows 2025 is not available yet until https://github.com/hashicorp/terraform-provider-azurerm/issues/31036 is done.

Last solution stop using Windows Node Pool in favor of Azure VM agents.…


dduportal:

PR #1465, opened 2026-05-21
fix(privatek8s-sponsored) add missing PE to reach our private ACR

dduportal:
I forgot this code block as it was hidden under the Kubernetes resources. It's required to ensure we get container images.

It will also (re)creates the windows 2022 node pool as #1464 failed to apply because the Azure API did not clean up the failed node pool with Generation 2 VM and orphaned it 😮‍💨 

Related to https://github.com/jenkins-infra/helpdesk/issues/5091#issuecomment-4381409952


dduportal:
`Plan: 5 to add, 0 to change, 0 to destroy.`

PR #1468, opened 2026-05-22
feat(infra.ci,release.ci,privatek8s-sponsored) add required resources for infra.ci and release.ci in the sponsored subscription

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5091#issuecomment-4381409952

Follows up #1458 (and its fixes from #1459 up to #1467 )

Requires azurevmagents module update to support many controller Azure IDs (#1469)

Along with the Terraform plan, I've opened #1471 in advance to check the impact of the removal  of CDF resources to ensure we don't forget anything here.

Blocked by #1469 and #1473 

Blocks #1474

PR #1469, opened 2026-05-22
feat!(modules/azure-jenkinsinfra-azurevm-agents) allow setting up many controller Azure identities instead of one and a set of additionals

dduportal:
When migrating controllers, we sometimes have an overlap between the new and old one. Both are expected to be allowed to spin up Azure VM agents.

The module already support additional identities, but it implies a set of destroy/add resources when cleaning up the old controller.

As such, the module is changed to support many Controller identities using a list of IDs. Using the `for_each` ensures that each resource is identified by a Key which is the Controller Identity ID so only destroy or removed IDs will happen.

Blocks #1468 

BREAKING CHANGE: Any usage of the module must be adapted to specify the plural form of the input variable (already done for our existing calls to the module)

----

Note: this change adds moved blocks to migrate existing resource internally in the Terraform state.
Once successfully applied, these blocks can be safely removed (most probably in #1468 )


dduportal:
- `Plan: 0 to add, 0 to change, 0 to destroy.`
- THe staging check fails due to the `moved` blocks

LGTM

PR #1473, opened 2026-05-22
fix(modules/azure-jenkinsinfra-azurevm-agents) avoid dynamic `for_each` structure in modules (in favor of static counts)

dduportal:
Fixup of #1469


dduportal:

PR #1474, opened 2026-05-22
feat(privatek8s-sponsored) add `kubernetes_*` resources

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5091#issuecomment-4381409952

Nothing specific all resources are new.

- (empty) Disks from sponsored RGs for azuredisk CSI PVs/PVCs
- UAID from sponsored RGs are used for federated identities

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -11,9 +11,9 @@ resource "local_file" "jenkins_infra_data_report" {
       },
     },
     "infra.ci.jenkins.io" = {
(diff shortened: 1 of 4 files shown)


dduportal:
`Plan: 21 to add, 0 to change, 1 to destroy.`

- 20 new resources
- The local report file is added and destroyed
🚀

PR #1471, opened 2026-05-22
cleanup(privatek8s,infra.ci,release.ci): decommission `kubernetes_*` and CDF resources (phase 1)

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5091#issuecomment-4554160116

First step to ensure no more resources depends on the `kubernetes` provider named `privatek8s`

PR #1481, opened 2026-05-27
cleanup(privatek8s) phase 2 removal: intermediate azureRM resources

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5091#issuecomment-4554160116

Follow up #1471

This intermediate step is required to ensure:

- Temp. resources used during migration are removed
- The lock on the public IP is removed
- The kubernetes provider for `privatek8s` and its associated outputs is removed


dduportal:
`Plan: 1 to add, 0 to change, 13 to destroy.`

PR #1491, opened 2026-06-03
fix(agent-2.trusted.ci.jenkins.io) increase data disk size to ensure more than 20% disk space is available

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5165


dduportal:

PR #1493, opened 2026-06-04
chore: replace `var.terratest` usage with `var.environment`

dduportal:
ref. https://github.com/jenkins-infra/helpdesk/issues/5169#issuecomment-4623739437


dduportal (commented):

PR #1521, opened 2026-08-26
feat(cert.ci.jenkins.io) set up commons resources without ACR resources (yet)

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5269

Need to re-run checks once https://github.com/jenkins-infra/azure-net/pull/559 is applied with success.

Note: might need to be applied twice to ensure all NSG rules are properly removed and then created.

PR #1523, opened 2026-08-27
hotfix(cert.ci.jenkins.io) ensure the correct "commons" subnet is used 

dduportal:
fixup of 1831e6c (#1521) where I mispelled the subnet name 

Ref. https://github.com/jenkins-infra/helpdesk/issues/5269


dduportal:
`Plan: 2 to add, 2 to change, 2 to destroy.`

Both the PE (subnet change) and the report are recreated as expected.

No expected impact as Docker engine will fallback to DockerHub if a ny use in cert.ci.

PR #1528, opened 2026-09-04
fix(cert.ci.jenkins.io) ensure the AzureVM agents UAID does not depend on the controller or agent components but on the 'commons' RG instead + cleanup

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5269#issuecomment-5541021978

⚠️ This PR will break the ability for cert.ci.jenkins.io agents to write into the build report storage. Pagerduty will most probably alert us in ~3h if no new agent health build is executed after changing the controller setup (see issue comment).

### DIFF
--- a/cert.ci.jenkins.io.tf
+++ b/cert.ci.jenkins.io.tf
@@ -90,17 +90,11 @@ module "cert_ci_jenkins_io_azurevm_agents_jenkins_sponsored" {
     privatevpn_subnet = data.azurerm_subnet.private_vnet_data_tier.address_prefixes
   }
 }
-resource "azurerm_resource_group" "cert_ci_jenkins_io_controller_jenkins_sponsored" {
-  provider = azurerm.jenkins-sponsored
-  name     = module.cert_ci_jenkins_io.controller_resourcegroup_name # Same name on both subscriptions
-  location = data.azurerm_virtual_network.cert_ci_jenkins_io_sponsored.location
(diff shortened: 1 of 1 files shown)


dduportal:
`Plan: 4 to add, 0 to change, 5 to destroy.`

- 1 destroy is the deleted RG for controller (will be recreated and fully managed by our Terraform module for https://github.com/jenkins-infra/helpdesk/issues/5089#issuecomment-5540956527)
- 1 add/destroy is the UAID
- 1 add/destroy is the report (where we'll retrieve the new ID of the new UAID)
- 1 add/destroy is the role assignment which allows the controller VM's UAID to assign the agents UAID to new VMs
- 1 add/destroy is the role assignment which allows the agents VM's UAID to read/write to the build reports storage

lemeurherve (approved):
> This PR will break the ability for cert.ci.jenkins.io agents to write into the build report storage

Does that mean https://github.com/jenkins-infra/helpdesk/issues/5123 should be reopened afterward?

Or is it just temporary?

dduportal:
> Does that mean https://github.com/jenkins-infra/helpdesk/issues/5123 should be reopened afterward?

Nope: we only need to update the cert.ci setup (puppet) with the new ID and we're good

dduportal:
See https://github.com/jenkins-infra/jenkins-infra/pull/5043 for the puppet update

PR #1533, opened 2026-09-05
cleanup(cert.ci.jenkins.io) remove setup of the removed CDF controller VM

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5089#issuecomment-5552132178


dduportal:
`Plan: 6 to add, 2 to change, 35 to destroy.`

- 2 changed resources: NSG rules from which the private IP of the CDF (to-be-destroyed) controller is removed
- 6 added / destroyed + 6 destroyed: all are role assignments in modules. Changed because of index change (could be improved but we don't really care honestly)
- All the other resources are destroyed resources

lemeurherve (approved):

PR #1539, opened 2026-09-11
chore: update all `kubernetes_*` resources to `kubernetes_*_v1` to remove deprecation warnings with Kubernetes provider 3.x

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/5083


dduportal:
`Plan: 82 to import, 0 to add, 11 to change, 0 to destroy.` almost there. Need to find why some resources are changed

dduportal:
> `Plan: 82 to import, 0 to add, 11 to change, 0 to destroy.` almost there. Need to find why some resources are changed

The 11 resources are the `kubernetes_secret_v1` resources which have their attribute [`[wait_for_service_account_token](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/secret#wait_for_service_account_token-1)`](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/secret#wait_for_service_account_token-1-) set to `true` as part of the re-import.

It looks like a weird but harmless bug in the provider, at least in the current 2.x version: it should not be the case for secrets of type `Opaque`.

Currently checking locally with the 3.x provider (as we did not see this behavior in the Terraform AWS project which is already on 3.x)

dduportal:
> Currently checking locally with the 3.x provider (as we did not see this behavior in the Terraform AWS project which is already on 3.x)

No changes. So let's roll

Checks are failing due to test unable to handle the `migrate.tf`. Subsequent PR will fix by removing it once applied

PR #1547, opened 2026-09-14
chore(locals) move definition of the `external_services` collection to a local YAML file

dduportal:
I need to add (and track) `census.jenkins.io`'s outbound IP for https://github.com/jenkins-infra/helpdesk/issues/4666.

I relaized that using a YAML file, like we did in jenkins-infra/terraform-aws-sponsorship, would be a good step here. Not strictly required so here is a simple PR with the existing data first.

- Same pattern as https://github.com/jenkins-infra/terraform-aws-sponsorship/pull/586
- Renamed and simplified the `updatecli`'s end_dates definition from a YAML local file to avoid confusion (as it's not the same file).
- Also updating the local `shared-tools` for hygiene (forgotten last week)

----

Expecting no changes


lemeurherve (approved):
> No changes. Your infrastructure matches the configuration.

PR #1548, opened 2026-09-14
feat(trusted.ci.jenkins.io) allow SSH access from controller to the remote census.jenkins.io VM

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4666

Blocks https://github.com/jenkins-infra/jenkins-infra/pull/5064

Note: the plan expects 2 changes:
- 1 outbound SSH (the one we need)
- But also an inbound on the JNLP port from census.jenkins.io: this one is harmless since trusted.ci.jenkins.io is NOT exposed publicly (no public IP).


dduportal:
`Plan: 0 to add, 2 to change, 0 to destroy.` as expected

